PR #57, opened 2026-06-24
fix: offer host update from a pre-release to its GA release

hdkshingala:
## Problem

A desktop on `1.0.0` whose host was installed from `1.0.0-rc.1` reported **"Up to date"** in Settings → Host and never auto-updated to the published `1.0.0`. Both the manual "Check for updates" button and the coordinated host auto-update were affected.

## Root cause

The host "update available?" decision in `buildUpdateState` (`host-management-ipc.ts`) reused `compareSemver` from `cli-discovery.ts`, whose parser strips the pre-release suffix (`v.split("-")[0]`). So `compareSemver("1.0.0-rc.1", "1.0.0") === 0` → `updateAvailable = false`.

That single flag gates **both** paths:

- Manual check → `refreshRegistryUpdateState` → `buildUpdateState`
- Launch/quit auto-update reconcile → `desktop-startup.ts` `if (result.updateAvailable …)`

The auto-update was wired correctly all along — it was starved by the wrong comparison. The strip is intentional for `compareSemver`'s *other* consumer (the CLI bundled-vs-PATH reconcile, where channel discrimination is out of scope); it just shouldn't drive host-update detection. The CLI's `traycer host update` already installs `manifest.latest` unconditionally, so only detection needed fixing.

## Fix

- New `compareHostVersions` in `cli-discovery.ts`: full SemVer precedence including pre-release ordering (`1.0.0-rc.1 < 1.0.0`, numeric vs alphanumeric identifiers, longer identifier list wins; build metadata ignored; returns `0` on unparseable input).
- `buildUpdateState` now uses it. `compareSemver` and its CLI-reconcile tests are left unchanged.
- The registry cache stores raw version strings and recomputes `updateAvailable` on read, so the fix lands on the next check/launch with no cache reset needed.

## Tests

- New `compare-host-versions.test.ts` — rc→GA, inter-pre-release ordering, no-downgrade, build metadata, unparseable input.
- `registry-update-cache.test.ts` — +2 cases (rc host + GA latest → update available; GA host + older rc latest → no downgrade).
- Verified green: `desktop:compile`, `desktop:lint`, `desktop:build`, format, and the affected test suites.


anur4ag:
@coderabbitai full review

PR #63, opened 2026-06-24
Update settings danger zone rows

hdkshingala:
## Summary
- Move file edit snapshots and local app state reset into the General settings danger zone.
- Rename the destructive local data actions to File Edit Snapshots and Local app state.
- Render the danger zone inline with normal settings rows instead of a nested rounded card.

## Verification
- `bun run test src/components/settings/panels/__tests__/general-settings-panel.test.tsx`
- `bun run compile`
- `bun run lint`
- `npx -y react-doctor@latest . --verbose --diff main --offline --no-score`
- `git diff --check`

Note: full `bun run react-doctor` reports pre-existing warnings outside this change; the changed-files check is clean.

### DIFF
--- a/clients/gui-app/src/components/settings/panels/__tests__/general-settings-panel.test.tsx
+++ b/clients/gui-app/src/components/settings/panels/__tests__/general-settings-panel.test.tsx
@@ -4,6 +4,7 @@ import {
   render,
   screen,
   waitFor,
+  within,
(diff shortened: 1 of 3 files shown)


hdkshingala:
@coderabbitai review

AmiteshwarRandhawa (approved):

PR #65, opened 2026-06-24
feat(gui-app): history-row right-click menu + fix open-in-new-window

anur4ag:
## Summary

Replaces the default OS/Electron right-click "Copy Link" on history rows (which copied the in-app href) with a custom Radix context menu, and fixes "Open in New Window" for an epic already open in the current window.

Applies to all three surfaces that share the history row: the History overlay, the `/epics` tab, and the home-embedded recent list.

## Changes

**Custom right-click menu**
- **Open in Background** — opens the epic tab behind the current surface without switching or closing it (new `openEpicTabInBackground` canvas action; no-op if already open). Hidden for **phase** rows, whose migration route (`migrationSource=phase`) a plain canvas tab can't carry.
- **Open in New Window** — routes through `useEpicOpenInNewWindowFlow`, so an epic already open **in this window** is popped out via the ownership-aware move flow, including its **unsynced-edits confirm dialog** — instead of silently focusing the current window (the bug). Epics open only elsewhere / nowhere, and all phases, delegate to `openEpicInNewWindow` (focus the owning window, else spawn a new one; phases keep `migrationSource=phase`). Hidden in web mode.

**Bug fixed**
Previously, right-clicking a history row whose epic was already open in the current window's tab strip and choosing "Open in New Window" did nothing — the action called `requestFocus(bridge.windowId)` on the already-focused window. It now pops the tab out to a new window.

**Cleanups / centralization**
- Removed the dead focus-self branch + `isEpicOpenInThisWindow` from `openEpicInNewWindow` (the open-here case is now handled by the flow).
- Extracted `createEpicViewTab` + `appendedEpicTabState` so `openEpicTab` and `openEpicTabInBackground` share one tab-append payload (~45 lines of duplication removed).

## Verification
- `bun run --filter @traycer-clients/gui-app compile` — clean
- ESLint + `react-doctor` on changed files — clean
- Tests: canvas store + panel + new-window action (46) and `components/layout` + `stores/tabs` regression net (116) — all pass. New tests cover menu gating (epic vs phase), move-flow routing (open-here / elsewhere / nowhere / phase), and the background action.

### DIFF
--- a/clients/gui-app/src/components/epic-tabs/__tests__/epic-tab-host.test.tsx
+++ b/clients/gui-app/src/components/epic-tabs/__tests__/epic-tab-host.test.tsx
@@ -422,12 +422,12 @@ describe("EpicTabHost keep-alive", () => {
 
     await waitFor(() => {
       expect(
-        useEpicCanvasStore.getState().firstTabIdForEpic("epic-a"),
+        useEpicCanvasStore.getState().resolveTabIdForEpic("epic-a"),
       ).not.toBeNull();
     });
     const firstCreatedTabId = useEpicCanvasStore
       .getState()
-      .firstTabIdForEpic("epic-a");
+      .resolveTabIdForEpic("epic-a");
     if (firstCreatedTabId === null) {
       throw new Error("expected first created tab");
     }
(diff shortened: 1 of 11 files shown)


anur4ag:
@coderabbitai full review

hdkshingala (approved):

PR #66, opened 2026-06-24
fix(stream): make host stream transport reliable across restart, reload, and sleep

hdkshingala:
## Problem

After the repo split, the local `make dev-desktop` stack surfaced two linked WebSocket stream-client bugs that made the app unusable in common flows:

1. **Reload skeleton** — on a UI reload the epic canvas, file tree, and terminals failed to load (infinite skeleton); only clearing tokens recovered it.
2. **Closed-client crash** — restarting the host from the Settings page crashed with *"Cannot subscribe with a closed WsStreamClient"* → "Something went wrong" error boundary.

Both traced to how stream clients were constructed and owned: a StrictMode-fragile module mirror (`liveHostStreamClient`) that could strand a closed/`null` client, plus an app-wide client that was rebuilt-and-closed on every endpoint change.

## Fix

Converge the epic stream onto the same **session-owned durable transport** model chat/terminal already use, make the app-wide client stable, and harden the shared transport.

- **Epic owns its durable transport** (`openOwnedDurableStreamClient`); delete the module mirror + the `handleStreamClients` reconcile the reload skeleton traced back to.
- **App-wide client keyed on host identity**, not the endpoint URL, with an immediate `reconnectAll` nudge on endpoint change — a host restart re-dials the same client instead of rebuilding-and-closing it and churning every consumer.
- **Endpoint-change re-dial generalized into `openDurableStreamTransport`** so the epic/chat/terminal session transports re-dial instantly on a Settings-page restart instead of waiting out the ~75s pong timeout on a half-open socket. Sleep/wake recovery (`online` + OS-resume) is preserved and now symmetric across both transport scopes.
- **Epic acquire effect gated on a non-null active host** — a transient host-null window now defers acquisition and recovers when the host binds, instead of the factory throwing into the root error boundary and tearing down the whole app.
- **`subscribe()` on a closed client degrades to an inert no-op session** (with a public `isClosed()` accessor) instead of throwing — defense-in-depth against a stale late subscribe.

## Review findings fixed (all severities)

A multi-agent review of the diff surfaced and adversarially verified these; all are addressed here:

| Severity | Finding |
|---|---|
| High | Epic factory threw on `activeHostId === null` → whole-app teardown via the root error boundary |
| Medium | Session transports lacked the endpoint-change nudge → up to ~75s stall on a Settings-page restart |
| Medium | Lost host-restart recovery test coverage (deleted transport-swap test) |
| Low | `isClosed()` orphaned after the deeper fix → wired as a defense-in-depth degrade |
| Low | StrictMode double-invoke of the nudge hook was untested |

## Testing

- `tsc -b` (gui-app + shared via project refs): clean
- ESLint (gui-app + shared changed files): clean
- Full suites: **gui-app 3113 passed / 1 skipped**, **shared 191 passed**
- New tests: durable-transport endpoint-redial, provider null-host gate, StrictMode nudge dedup, closed-client inert degrade
- Live-verified on `make dev-desktop`: reload + Settings-restart both recover the canvas with no crash


anur4ag:
@coderabbitai full review

anur4ag (approved):

PR #70, opened 2026-06-25
fix(desktop): user-initiated updates, restart confirm, and macOS relocation handling

hdkshingala:
## Problem

The desktop auto-updater downloaded any available update on launch and, the
moment the download finished, prompted to restart — mid-work, regardless of
focus. On macOS a copy run from the mounted `.dmg` / a Gatekeeper
App-Translocation path sits on a **read-only volume**, so Squirrel.Mac can't
apply the update; that failure was swallowed (status `"ready"` short-circuited
the error handler), so **"Restart" did nothing, with no feedback**.

## Fix

Make the update path user-driven and legible, and handle the macOS read-only
location the platform-standard way (move-to-Applications, like VS Code / Slack).

- **No auto-download.** `update-available` surfaces a new `"available"` state.
  The user starts the download from a compact header control (mic-style progress
  ring) or an in-app **"Update available"** toast with a Download action.
  Progress is tracked as whole-percent.
- **Confirm before restart.** When ready, a shared blur-backdrop modal confirms
  the restart (running agents are interrupted) before `quitAndInstall`.
- **Focus-aware OS notification.** Fired only when no window is focused; in-app
  surfaces never nag while the user is in the app.
- **macOS read-only / App-Translocation.** A move-to-Applications prompt is
  offered **once** (decline persisted), **after the window exists** (no
  boot-blocking parentless modal), and only on a packaged build with a real
  update feed. While blocked, the download/restart affordances are disabled with
  an explanation and the OS notification points at the fix.
- **Surface install failures.** A `quitAndInstall` error (e.g. read-only volume)
  is no longer swallowed by the `"ready"` guard — it clears the install flag,
  emits an error, auto-closes the restart modal, and (during the quit-time host
  reconcile) keeps the app open so the error is readable.

## Cross-platform

The relocation flow is **macOS-only** and inert on Windows/Linux (no
Applications folder): `isUpdateBlockedByLocation()` gates on `darwin`, so there
is no prompt, no disabled button, and no blocked toast there. The read-only →
"move to Applications" error message is also gated to macOS; off macOS a
read-only/permission failure falls through to the generic install message.

## Review findings fixed

A max-effort multi-agent review surfaced and adversarially verified **14
findings**; all are addressed: stranded restart-modal store, every-launch
relocation nag, silent trashing of an existing `/Applications` copy,
boot-blocking pre-window modal, OS-notification/ready-button ignoring the blocked
state, frozen blocked-reason, missing update-feed gate, quit-time reconcile
force-quit after a failed install, error-toast replay suppression, narrow
read-only hint list, persistent blocked-toast nag, and copy/predicate
duplication.

## Testing

- `bun run compile` (desktop + gui-app): clean
- ESLint + `react-doctor` (changed files): clean
- Suites: **desktop 350 passed**, **gui-app 3127 passed / 1 skipped**
- New tests: relocation prompt (ask-once, feed gate, async accept/decline,
  conflict), install-failure surfacing, blocked-state UI, off-macOS message gate


anur4ag:
@coderabbitai full review

PR #71, opened 2026-06-25
fix(gui-app): settle in-flight attachment reads on epic dispose

YuriPerro:
## Summary

`readAttachmentBytes` (in `clients/gui-app/src/stores/epics/open-epic/store.ts`) parks a waiter that `observe`s the live Y.Doc `attachments` map and resolves once the bytes sync in. A waiter is torn down only by its own resolve or by the caller's abort signal.

`dispose()` tore down the projector, stream client, and replica (`destroyReplica(doc, awareness)`) but never drained `attachmentReadWaiters`. So when a read is still in flight at dispose time — e.g. a session disposed by the registry's MRU prune while an image is still streaming in (the method has no fixed give-up), where the caller's `AbortSignal` isn't guaranteed to fire — the waiter is left:

- with a `.observe(onChange)` still registered on the now-`destroy()`ed doc's `attachments` map, and
- a promise that never resolves (orphaned for the page lifetime).

This drains the waiter set in `dispose()` before the replica is destroyed: each pending read is settled to `null` and its observer unbound from the live map. `settle` is hoisted onto the waiter so `dispose` and the abort path share one teardown. The abort and resolve paths are unchanged.

Verified with a new test (added next to the other dispose tests): park a waiter on an unsynced hash, dispose without firing the abort signal, and assert the promise resolves to `null`. It fails before the fix (promise never settles) and passes after.

## Related issue

n/a

## Checklist

- [ ] Pre-commit hooks pass — not run locally (pre-commit not installed); lint + format pass and the diff is plain TS

### DIFF
--- a/clients/gui-app/src/stores/epics/open-epic/__tests__/store.test.ts
+++ b/clients/gui-app/src/stores/epics/open-epic/__tests__/store.test.ts
@@ -2281,4 +2281,37 @@ describe("createOpenEpicStore", () => {
       opened.dispose();
     });
   });
+
+  it("settles an in-flight attachment read when disposed without an abort", async () => {
+    const { factory } = fakeFactory();
+    const opened = createOpenEpicStore({
+      epicId: "epic-a",
+      streamClientFactory: factory,
+      userId: null,
+      onAuthError: null,
+    });
+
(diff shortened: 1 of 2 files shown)


anur4ag:
LGTM, thanks for fixing!

PR #73, opened 2026-06-25
fix(gui-app): label report-issue dialog fields for screen readers

YuriPerro:
## Summary

The `Field` wrapper in the desktop Report-an-Issue dialog (`clients/gui-app/src/components/layout/dialogs/desktop/report-issue-dialog.tsx`) rendered a Radix `<Label>` with no `htmlFor` and the control as a sibling child, not nested. Radix `Label` associates only via `htmlFor`→`id` or by wrapping the control, so none of the five inputs (Title, What happened?, Steps to reproduce, Expected behavior, Actual behavior) had a programmatic label — only a placeholder.

Result: screen readers announce no name for any field (WCAG 1.3.1 / 4.1.2), and clicking a visible label doesn't focus its input. This form is opened from the command palette, the app menu, and update-error toasts.

`Field` now owns the association: it generates an id with `useId()`, sets it on the label's `htmlFor`, and threads it onto its single control child. Fixes all five fields with no call-site changes.

Added a test that renders the dialog and asserts each field is reachable by its accessible name (`queryByLabelText`); it fails before the change and passes after.

## Related issue

n/a

## Checklist

- [ ] Pre-commit hooks pass — not run locally (pre-commit not installed); lint + format pass and the diff is plain TS

### DIFF
--- a/clients/gui-app/src/components/layout/dialogs/desktop/__tests__/report-issue-dialog.test.tsx
+++ b/clients/gui-app/src/components/layout/dialogs/desktop/__tests__/report-issue-dialog.test.tsx
@@ -1,0 +1,47 @@
+import "../../../../../../__tests__/test-browser-apis";
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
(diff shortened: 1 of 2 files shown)


YuriPerro:
Withdrawing this — on reflection it's too minor to be worth the review overhead. Thanks!

PR #82, opened 2026-06-26
fix(protocol): version host.getRateLimitUsage accountContext as v1.1

Sahil1337:
accountContext was added in place on v1.0, a silent contract break: the handshake compares version numbers only, so a new client and a released v1.0 host both report "1.0", yet the old host's strict schema rejects the unknown field at request time (RPC_ERROR -> "usage unavailable").

- Revert v1.0 to the empty request and make it NON-strict, so a new client can strip-project its request down to {} for a released v1.0 host (the client-side downgrade in ws-rpc-client uses safeParse, which strips only when the schema is non-strict).
- Ship accountContext as a new v1.1 minor with a v1.0 -> v1.1 upgrade bridge that fills the default (PERSONAL) account for old clients.
- Canonical request type points at v1.1; no unversioned schema export.

## Checklist

- [ ] `bun run build`, `bun run lint`, and `bun run test` pass
- [ ] Code is formatted (`bun run format`)
- [ ] Pre-commit hooks pass (`pre-commit run --all-files`)
- [ ] Tests added/updated where it makes sense
- [ ] Commits are signed off (`git commit -s`) per the [DCO](../CONTRIBUTING.md#developer-certificate-of-origin-dco)

### DIFF
--- a/protocol/src/host/rate-limit/contracts.ts
+++ b/protocol/src/host/rate-limit/contracts.ts
@@ -1,12 +1,42 @@
-import { defineRpcContract } from "@traycer/protocol/framework/index";
 import {
-  rateLimitUsageRequestSchema,
+  defineRpcContract,
+  defineUpgradePath,
+} from "@traycer/protocol/framework/index";
+import { DEFAULT_ACCOUNT_CONTEXT } from "@traycer/protocol/common/schemas";
+import {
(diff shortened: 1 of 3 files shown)


hdkshingala:
Handled in #84

PR #93, opened 2026-06-27
Add Gemini ACP harness protocol support

hdkshingala:
## Summary
- Add Gemini to the latest harness/provider protocol schemas while keeping v1.0 frozen and filtering both Grok and Gemini through the existing v2-to-v1 downgrade bridges.
- Add Gemini ACP runtime and persisted chat-session anchor schemas.
- Wire Gemini GUI surface copy/icons/onboarding/provider settings, including bare-command OAuth support for `gemini` and a focused reauth banner test.

## Verification
- `bun run --filter @traycer/protocol compile`
- `bun run --filter @traycer-clients/gui-app compile`
- `bun run test src/host/agent/gui/__tests__/grok-versioning.test.ts` in `protocol/`
- `bun run test src/components/chat/composer/__tests__/provider-reauth-banner.test.tsx` in `clients/gui-app/`

### DIFF
--- a/clients/gui-app/src/components/chat/composer/__tests__/provider-reauth-banner.test.tsx
+++ b/clients/gui-app/src/components/chat/composer/__tests__/provider-reauth-banner.test.tsx
@@ -119,7 +119,7 @@ function claudeState(
 function geminiState(): ProviderCliState {
   return {
     ...claudeState({
-      oauthArgs: [],
+      oauthArgs: ["--list-sessions"],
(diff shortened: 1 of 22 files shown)


hdkshingala:
Google is sunsetting Gemini CLI and transitioning to Antigravity CLI: https://github.com/google-gemini/gemini-cli/discussions/28017

Antigravity doesn't support ACP, so we will see in future when the support is added.

PR #106, opened 2026-06-27
fix(auth): refresh token on device wake so long sessions survive sleep

### LINKED ISSUE #103: Long running tasks fail due to traycer auth
### Component

Desktop app

### Version

1.0.0

### OS / platform

macOS

### What happened?

Hey guys traycer doesnt seem to refresh auth token in the background. So long running tasks are unable to proceed and launch sub agents. 

Whn i come back online, I am able to say retry and it works

Image

### Steps to reproduce

1. Run a log running command which spawns agents
2. after a couple hours task will no longer be able to spawn as it gets blocked

### Logs / output

```shell

```

ghostyfreak:
## Summary

The proactive refresh scheduler added in #83 arms a `setTimeout` whose monotonic countdown freezes while the OS is suspended, so a session that sleeps through the ~4h token TTL wakes with a dead bearer and the next live host call (a resumed long-running task spawning a sub-agent) 401s — only the reactive 401 path recovers, after a user-visible failure. This adds `ProactiveRefreshScheduler.notifyResumed()` to re-evaluate the refresh deadline against the wall clock immediately, and wires the renderer's `AuthService` to the OS-wake signals (`onSystemResumed` + `window 'online'`) the stream transports already use, so on wake the bearer refreshes and propagates to the host in place via the existing `credentialUpdate` push before that first call. Tests cover the new scheduler hook and the AuthService wake wiring.

## Related issue

Closes #103

## Checklist

- [ ] Pre-commit hooks pass (`pre-commit run --all-files`) — `pre-commit` not installed in this environment

### DIFF
--- a/clients/gui-app/src/lib/auth/__tests__/auth-service.test.ts
+++ b/clients/gui-app/src/lib/auth/__tests__/auth-service.test.ts
@@ -194,6 +194,26 @@ function status(code: number): Promise<Response> {
   return Promise.resolve(new Response(null, { status: code }));
 }
 
+function base64url(value: string): string {
+  return btoa(value).replace(/\+/g, "-").replace(/\//g, "_").replace(/=+$/, "");
+}
+
+/**
+ * A JWS-shaped access token whose payload carries a decodable `exp`, so the
+ * proactive refresh scheduler arms off it (the arbitrary opaque strings the
+ * other tests use carry no `exp` and leave the scheduler disabled).
+ */
+function jwtExpiringInMs(fromNowMs: number): string {
+  const header = base64url(JSON.stringify({ alg: "RS256", typ: "JWT" }));
+  const payload = base64url(
+    JSON.stringify({
+      id: "user-1",
+      exp: Math.trunc((Date.now() + fromNowMs) / 1000),
+    }),
+  );
+  return `${header}.${payload}.signature`;
+}
+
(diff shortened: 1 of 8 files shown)


anur4ag:
@ghostyfreak please run pre-commit on your changes

ghostyfreak:
@anur4ag done — ran pre-commit and pushed the fix; all checks are green now.

The failure was in the `workspace-checks` (format) hook, not my auth changes. Touching `clients/shared` invalidated that project's nx format cache, so CI re-ran prettier and collapsed four pre-existing multi-line union types onto single lines (`auth-validation.ts`, `i-stream-session.ts`, `ws-stream-factory.ts`, `runner-host.ts`). I committed exactly that formatting in `style(shared): collapse short union types…` — pure formatting, no logic change.

Heads-up on the root cause: the repo pins `prettier: "^3.8.4"`, but CI resolves the newly-published **3.8.5**, whose union formatting differs — so any PR that touches a cached project will surface latent reformats like this. Might be worth pinning prettier to an exact version. Happy to open a separate PR for that if useful.

hdkshingala (approved):

hdkshingala:
Thank you for the contribution and finding the prettier issue, @ghostyfreak!

We will be performing an all package version bump soon, so we will handle it during the same.

PR #116, opened 2026-06-28
fix(auth): re-read/adopt on reject in shared bearer revalidator

tanveergill:
Part of the **auth refresh-token grace window** work (client-side hardening; ticket `fix1-reject-adopt`).

When a refresh is rejected but a sibling process already rotated and persisted a fresh token, the shared `createBearerRevalidator` now re-reads the store (a short bounded poll) and **adopts** the sibling's token instead of concluding `rejected` — closing a spurious-failure window for racing refreshers (CLI monitor + one-shot commands) whose refresh lands outside the server-side grace window.

- `clients/shared/auth/bearer-revalidator.ts`: reject-path bounded poll + adopt (injected `delay`; `REJECT_REREAD_POLL_INTERVAL_MS=100`, `REJECT_REREAD_ATTEMPTS=5`).
- CLI call sites (`host-rpc.ts`, `monitor.ts`) pass the delay factory.
- GUI intentionally unchanged — it has its own identity-safe guard (`signOutIfRefreshCredentialDead`); the GUI must not blind-adopt (shared multi-window token slot).

Tests: 12 bearer-revalidator + 82 auth suite green; eslint clean.

### DIFF
--- a/clients/shared/auth/__tests__/bearer-revalidator.test.ts
+++ b/clients/shared/auth/__tests__/bearer-revalidator.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import {
   createBearerRevalidator,
+  REJECT_REREAD_ATTEMPTS,
   rotateAndPersistBearer,
   type BearerStore,
 } from "../bearer-revalidator";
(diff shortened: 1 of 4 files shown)


tanveergill:
Superseded by the combined PR #119 (consolidated to one submodule PR per request). Same commit, DCO-signed, now on traycer/traycer-authn-refresh-hardening.

PR #118, opened 2026-06-28
auth: map authn refresh 409 to network-error in refreshAuthTokenViaHttp

tanveergill:
Client companion to the **authn refresh-token grace window** server change (ticket `c-grace-window`, client half; pairs with the authn-v3 PR in traycerai/traycer-internal).

The new authn grace window returns **HTTP 409** while a concurrent refresh is mid-rotation (the loser should retry, not sign out). This maps `409` → `network-error` in `refreshAuthTokenViaHttp` explicitly — it previously fell through the `status >= 300` catch-all, so this pins the mapping as intentional and tested. `401` → `rejected` / success → `refreshed` unchanged. Backward-compatible.

Tests: shared/auth 9/9; eslint clean.

### DIFF
--- a/clients/shared/auth/__tests__/auth-validation.test.ts
+++ b/clients/shared/auth/__tests__/auth-validation.test.ts
@@ -12,6 +12,7 @@
  */
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
+  refreshAuthTokenViaHttp,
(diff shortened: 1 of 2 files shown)


tanveergill:
Superseded by the combined PR #119 (consolidated to one submodule PR per request). Same commit, DCO-signed, now on traycer/traycer-authn-refresh-hardening.

PR #117, opened 2026-06-28
fix(gui-auth): re-provision full token pair to CLI on reactive refresh

tanveergill:
Part of the **auth refresh-token grace window** work (client-side; ticket `fix2-reactive-reprovision`).

The GUI re-seeds the CLI credentials file, but on a **reactive** refresh only the bearer was re-seeded (refresh token left stale) — so the file could hold a fresh bearer beside a spent refresh token. This adds `ensureLocalProvisioning(token, refreshToken)` on the reactive rotate path (`revalidateCurrentContextOnce`), mirroring the proactive path, so the file always carries the full rotated pair.

- `clients/gui-app/src/lib/auth/auth-service.ts`: full-pair re-provision on the same-user reactive rotate branch.
- Seeder dedupe (`CliCredentialSeeder`) confirmed unaffected.

Tests: gui-app auth suite 44/44 (2 new + 42 existing); eslint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #119, opened 2026-06-28
auth: refresh-token grace-window client hardening (reject-adopt + reactive re-provision + 409 mapping)

tanveergill:
Consolidated **auth refresh-token grace window** client-side work — three independent changes (zero file overlap) on one branch.

### 1. Reject re-read/adopt — `clients/shared/auth/bearer-revalidator.ts`
On a rejected refresh, bounded-poll the store and **adopt** a sibling's freshly-rotated token instead of concluding `rejected` — closes a spurious-failure window for racing refreshers (CLI monitor + one-shot commands) whose refresh lands outside the server grace window. CLI call sites pass the injected delay. GUI intentionally unchanged (it has its own identity-safe `signOutIfRefreshCredentialDead`).

### 2. Reactive full-pair re-provision — `clients/gui-app/.../auth-service.ts`
On a **reactive** refresh the GUI now re-provisions the FULL pair (`ensureLocalProvisioning(token, refreshToken)`) to the CLI credentials file, mirroring the proactive path — so the file never holds a fresh bearer beside a spent refresh token.

### 3. 409 → network-error mapping — `clients/shared/auth/auth-validation.ts`
Companion to the authn grace-window server change (traycerai/traycer-internal#4144): maps the new HTTP 409 (concurrent refresh mid-rotation) to `network-error` so a racing loser retries instead of signing out.

Verified together: gui-app + traycer-cli compile clean; shared/auth 21 + gui-app/auth 44 tests green; all three commits DCO-signed.

Supersedes the per-ticket PRs #116 / #117 / #118.

### DIFF
--- a/clients/gui-app/src/lib/auth/__tests__/auth-service.test.ts
+++ b/clients/gui-app/src/lib/auth/__tests__/auth-service.test.ts
@@ -1340,7 +1340,11 @@ describe("AuthService", () => {
       const service = trackService(new AuthService({ runnerHost: host }));
       await service.start();
       await service.signIn();
-      host.emitAuthCallback({ code: "old-token" });
+      host.deviceFlow.emitResult({
+        kind: "authorized",
+        token: "old-token",
+        refreshToken: "old-token-refresh",
+      });
       await vi.waitFor(() => {
         expect(useAuthStore.getState().status).toBe("signed-in");
       });
--- a/clients/gui-app/src/lib/auth/__tests__/auth-service.test.ts
+++ b/clients/gui-app/src/lib/auth/__tests__/auth-service.test.ts
@@ -1324,5 +1324,108 @@ describe("AuthService", () => {
       expect(hostB.openedExternalLinks).toHaveLength(0);
       expect(hostB.deviceFlow.startCalls).toBe(0);
     });
+
+    it("re-provisions the FULL pair to the local CLI on a reactive refresh", async () => {
+      const cli = new MockTraycerCli();
+      const host = new MockRunnerHost({
+        signInUrl:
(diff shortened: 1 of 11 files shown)


tanveergill:
Addressed CodeRabbit's auth-service finding (post-sign-out reprovision) in `dc1eb0d`:

- Provision and deprovision now serialize on a single chain (`enqueueCredentialOp`), so a provisioning retry can never interleave with — and land after — a concurrent sign-out's `cliLogout`.
- Each provision re-checks staleness once it holds the chain and skips the write if the session no longer matches the rotated token (`currentBearer !== token`, which `applySignedOut()` nulls). Sign-in keeps an `isDisposed`-only guard since it provisions *before* `currentBearer` is set.

Verified: gui-app compile clean, auth-service suite 44/44, eslint clean.

tanveergill:
Addressed CodeRabbit's identity-gate finding in `4857ba1`:

- `BearerStore.read()` now surfaces the persisted `userId` (new `StoredBearer` type; `cliBearerStore.read()` returns `stored.user.id`).
- **All three** adopt sites in `createBearerRevalidator` — before-refresh (STEP-1), the reject re-read loop (the flagged one), and the post-refresh sibling-rotated check (STEP-3) — now gate adoption on `persisted.userId === lease.identity.userId`. On mismatch we don't adopt. This mirrors the GUI `AuthService`'s deliberate refusal to blind-adopt the shared shell token.
- Added two identity-mismatch tests (a different account's token is not adopted, at STEP-1 and on the reject re-read).

Verified: cli + gui-app compile clean; shared/auth 23/23, gui-app/auth 48/48, cli monitor+host-rpc 12/12; eslint clean. DCO-signed.

PR #127, opened 2026-06-28
feat(gui,protocol): render a backgrounded subagent's live progress across the turn boundary

hdkshingala:
## Summary

Renders a backgrounded Claude subagent's live progress across the turn boundary — the **GUI + protocol** half of the background-task auto-resume feature. Pairs with the host-side change in `traycer-internal`.

- **protocol / accumulator**: on a **clean** `turn.completed`, keep a still-`streaming` subagent block streaming instead of finalizing it. A synchronous subagent's result returns to the model *before* the model finishes, so a subagent still running at a clean turn end is backgrounded and outlives the turn. `turn.stopped` / `turn.interrupted` still finalize it.
- **gui / chat store**: route a detached subagent's events (`subagent.*` by `blockId`; nested tool/file events by `parentBlockId`) to the **settled** message that owns its card, so updates after the foreground turn ends aren't dropped or mis-applied to a later turn. Falls back to active-turn routing when no message owns the block.

No wire-format change; both behaviors use existing schema/events.

## Validation

- `gui-app` compile green · accumulator tests **67/67**.
- Confirmed on the real host (`make dev-desktop`): the subagent card shows live progress through the background wait and finalizes on completion; no stray `<task-notification>` bubble on reload.

## Open review follow-ups (draft — to address in next commits)

From the focused review (`reviews/bg-subagent-realhost-fixes-review`):

1. **(P1) Degraded `turn.completed` preserves a running subagent.** The accumulator split currently treats *every* `turn.completed` as clean; degraded endings (e.g. `max_tokens`) should **not** preserve — parity with the host's `degradedStopReason(...) === undefined` keep-alive gate. Key the preserve on *clean* completion. — `protocol/src/host/agent/gui/agent-runtime-accumulator.ts`
2. **(minor) Detached updates advance the settled row `timestamp`.** The reducer rewrites `timestamp` to each detached event's timestamp, which moves the turn's completed-at / cache token until a snapshot repairs it. — `clients/gui-app/src/stores/chats/chat-session-store.ts`
3. **(minor) Transient live-row materialization force-finalizes streaming subagents** (`finalizeStreamingActionBlocks(..., "interrupted")`), a brief window where the card can flicker "interrupted" despite the terminal delta preserving it. — `clients/gui-app/src/stores/chats/chat-session-store.ts`

> Draft: review incorporation and UX-side changes land in follow-up commits.


hdkshingala:
@coderabbitai resolve

hdkshingala:
@coderabbitai review

PR #147, opened 2026-06-30
feat: progressive provider loading — return pending immediately, poll until settled

hdkshingala:
## Summary

- **Root cause**: On a fresh launchd boot, the interactive-shell PTY probe (PATH discovery + env-sourced API-key lookup) blocks for up to ~86 s. The `providers.list` RPC waited on this via two `Promise.all` barriers, leaving the GUI model picker and Settings→Providers panel visually empty for 2-3 min on first launch.

- **Fix**: Per-provider progressive loading — providers return `availabilityPending: true` immediately while shell-env probes run in the background; the client polls at 800 ms until all pending flags clear.

## Protocol changes

- `providerCliStateBaseShape` gains `availabilityPending: z.boolean().catch(false)` — old hosts that omit the field default to `false` (no regression).
- Froze the pre-v2.0 shape as `providerCliStateBaseShapeV10` (strict object); exported `providerCliStateSchemaV20`.
- `downgradeProviderCliStateV20ToV10` now strips `availabilityPending` before the strict V10 parse so old (v1.x) clients don't silently lose all providers.
- Added `upgradeProviderCliStateV10ToV20` / `upgradeProviderStateFromV10` / `upgradeProviderStateListFromV10`; updated all 11 `providers*UpgradeV1ToV2` functions from identity stubs to proper state mapping.
- Added `availabilityPending` to both `GuiHarnessOption` and `TuiHarnessOption` schemas with `.catch(false)`.

## GUI client changes

- **Poll predicate** (`use-providers-list-query`, `use-tab-providers-list-query`): fast-polls at 800 ms while any provider has `availabilityPending: true`.
- **Rail visibility** (`harness-rail-providers.ts`): `railHarnessVisible` includes `availabilityPending` so pending providers appear immediately.
- **Rail button UI** (`harness-model-picker-group.tsx`): pending button shows the provider icon at 25% opacity with a centered spinner overlay; `aria-label` says "loading…"; `aria-describedby` suppressed while pending; `railButtonTitle()` helper replaces a nested ternary (lint).
- **⌘+digit dispatch** (`use-picker-leader-scope.ts`): skips a pending button — mirrors the disabled state the rail renders.
- **Selection reroute guard** (`composer-toolbar-store.ts`): `effectiveSelectionFromHarnesses` now early-returns on a pending harness instead of rerouting — cold-boot would otherwise immediately switch the user off their selected provider.

## Test plan

- [ ] Fresh install: model picker and Settings→Providers now populate immediately with spinner overlays on pending providers; providers settle within seconds as the PTY probe completes.
- [ ] Old host (v1.x) connected to new client: `availabilityPending` omitted by host → defaults to `false` → old behavior, no blank providers.
- [ ] New host connected to old client (v1.x): V20→V10 downgrade strips `availabilityPending` before the strict parse → all providers visible.
- [ ] ⌘+digit on a pending provider: no-op (does not dispatch selection).
- [ ] Selected provider is still pending on cold boot: selection is preserved, not rerouted.
- All 450 protocol tests pass; all 3317+ GUI tests pass.


hdkshingala:
@coderabbitai review

PR #159, opened 2026-07-01
feat(gui): create child chats through the shared new-conversation modal

anur4ag:
## Problem

The chats-panel **"+"** opens the New Conversation modal (compose-first), but the per-row **"+" → New chat** eagerly created an empty child chat *tile* instead — two different UXs for the same "start a chat" action.

## Change

Route the per-row **New chat** action through the **same** modal, seeded as a child of that row.

- The modal open-request carries a `parentId` (`null` for top-level). The body threads it into `epic.createChat` (and the terminal path) so submit creates a **child**, and seeds the workspace from the **parent chat's binding** (via `useOwnerWorkspaceInheritanceSeed` on the active host) so the child **inherits the parent's worktree** — preserving the behavior of the eager path it replaces.
- In child mode the modal shows a **muted subtext naming the parent chat** and **hides the chat/terminal switcher** (child creation is chat-only here; the row's own dropdown still offers a child terminal agent).
- Removes the row's now-dead optimistic create / pending-placeholder / tile-open machinery (and the obsolete `buildChildChatCreateInput` + its test). The terminal-agent path and the row's host resolution are unchanged.

All other open-request callers (⌘K palette, in-pane openers) pass `parentId: null` and keep creating top-level chats.

### Scoping notes
- Child chats are created on the **active host** (like the chats-panel `+`). A reachable *remote*-host parent falls back to an empty, user-adjustable workspace rather than the remote binding.
- The row no longer renders an optimistic pending child placeholder for chats — the modal owns that UX now.

## Testing

- `bun run compile` — clean · `bun run lint` (changed files) — clean · `react-doctor --scope changed` — no issues
- `bun run test src/components/epic-canvas/sidebar src/lib/commands src/components/chat` — **665 passed** (updated the open-request shape assertions to include `parentId`)
- `pre-commit run` — all hooks pass (incl. nx-affected build/compile/lint/format) + DCO sign-off

Not visually verified in a running app (needs the local dev stack).

_Follow-up to #158 (the scroll fix), rebased onto `main` after that merged._

### DIFF
--- a/clients/gui-app/src/components/epic-canvas/sidebar/__tests__/chat-row-child-create.test.ts
+++ b/clients/gui-app/src/components/epic-canvas/sidebar/__tests__/chat-row-child-create.test.ts
@@ -1,6 +1,5 @@
 import { describe, expect, it } from "vitest";
 import type { WorktreeBindingEntry } from "@traycer/protocol/host/worktree-schemas";
-import { buildChildChatCreateInput } from "../use-chat-row-child-create";
 import { buildOwnerWorkspaceInheritanceSeed } from "@/lib/worktree/owner-workspace-inheritance-seed";
 import { buildFixedHostWorkspaceControlsScope } from "@/components/home/host-workspace-selector/host-workspace-controls-scope";
 import { resolveOwnerWorkspaceInheritanceSeed } from "@/hooks/worktree/use-owner-workspace-inheritance-seed";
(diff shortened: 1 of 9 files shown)


anur4ag:
Superseded by #160 — moved this change onto its own branch (`traycer/child-chat-modal`) branched from `main`, instead of the already-merged `fix-scroll-fix-new-chat` branch (#158).

PR #163, opened 2026-07-01
fix(host): stop cli-lock wedging forever on PID reuse, gate host update on busy check

hdkshingala:
## Summary

A user reported `traycer host update` permanently failing with `another traycer CLI mutation is in progress (holder.pid=..., reason=host-update, ...)`. Killing every Traycer process didn't clear it; deleting `~/.traycer/cli/.lock` by hand did.

Root cause: `cli-lock.ts`'s staleness check uses `process.kill(pid, 0)` to decide if a lock holder is still alive, but that only proves *some* process currently owns that PID — not that it's the original holder. Once the holder died, a recycled PID (same-user reuse read as alive silently, or different-user reuse hitting the existing `EPERM` "be conservative" branch) got treated as alive forever, wedging the lock with no self-recovery.

- Add an age-ceiling backstop (`MAX_LOCK_AGE_MS`, 10 min) that force-breaks a lock regardless of what the PID-liveness check concludes.
- Add a per-acquisition token to the lock metadata so `release()` does compare-and-delete instead of an unconditional unlink, closing a related torn-lock hazard (a false-positive break could otherwise let the original, still-alive holder delete a legitimate new holder's lock out from under it when it finishes). A missing token (pre-fix lock format) falls back to the old unconditional unlink for cross-version compatibility.
- Separately, `host update` was the one lock-guarded command that never checked whether the host had work in progress before restarting it (unlike `provisionHost`'s `assertHostNotBusy` gate used by install/uninstall/service-cycle). It now runs the same busy check, skippable with a new `--force` flag matching `host ensure`.

## Test plan

- [x] `bunx nx run @traycer-clients/traycer-cli:compile`
- [x] `bunx nx run @traycer-clients/traycer-cli:lint`
- [x] `bunx nx run @traycer-clients/traycer-cli:test` (347 tests, including 16 new: `store/__tests__/cli-lock.test.ts` and `commands/__tests__/host-update.test.ts`)
- [x] New regression test reproduces the exact reported failure mode: a lock with an alive `pid` but a `startedAt` past the age ceiling is broken instead of hanging until `waitMs`

### DIFF
--- a/clients/traycer-cli/src/commands/__tests__/host-update.test.ts
+++ b/clients/traycer-cli/src/commands/__tests__/host-update.test.ts
@@ -103,7 +103,9 @@ function fakeCtx(): CommandContext {
 
 describe("buildHostUpdateCommand busy-check gating", () => {
   afterEach(() => {
-    vi.clearAllMocks();
+    // resetAllMocks (not clearAllMocks) so a mockResolvedValue/
+    // mockRejectedValue configured in one test can't leak into the next.
+    vi.resetAllMocks();
   });
 
   it("rejects with E_HOST_BUSY without --force, and never calls installHost", async () => {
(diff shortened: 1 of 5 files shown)


hdkshingala:
@coderabbitai review

PR #195, opened 2026-07-04
fix(gui-app,desktop): eliminate phantom "New tab" on terminal-agent launch

AmiteshwarRandhawa:
## Problem

When a terminal agent launched into a fresh epic, a phantom "New tab" was left behind. Two independent causes:

1. **Main-process self-echo (desktop).** A window authors its own per-window state, but main bounced that write straight back to the originating window. The renderer debounces outbound writes, so an echo in flight could land **after** a newer local edit and clobber it — resurrecting a landing draft the window just closed.
2. **Blank-tab stacking (gui-app).** An empty epic seeds a blank "New tab" placeholder while the agent tile loads. When the agent tile then opened via `openTile`, it appended a second tab beside the blank instead of replacing it.

## Fix

- **desktop / `per-window-state-ipc.ts`:** while applying a window's own `perWindowState.update`, suppress the `change` push back to that same window. Main still persists the update and still pushes MAIN-initiated changes (restore, move-tab), which carry no origin. Suppression is tracked in a per-window `Set` so overlapping updates from two windows can't clobber each other's flag if the update path ever becomes async.
- **gui-app / `canvas/actions.ts`:** a permanent `openTile` while the active tab is a blank "New tab" now replaces that blank in place (browser new-tab semantics, mirroring `openTileInPane`) instead of stacking. Preview opens still append.

## Incidental

- `rendered-messages.test.tsx`: added the `coveredUntilMessageId` field that #187 made required on codex session anchors but missed in this fixture — the gui-app type-check was already red on `main` without it. Isolated in its own commit.

## Tests

- New regression tests for both the no-self-echo behavior and the fill-in-place behavior.
- `bun run compile` clean across desktop + gui-app; `runner-ipc` suite passes.


AmiteshwarRandhawa:
@coderabbitai resolve all

PR #205, opened 2026-07-06
fix(protocol): persist wakeup triggers additively for v1.1.x compat

hdkshingala:
## Summary
- A fired ScheduleWakeup was persisted inline as `kind: "wakeup"` in `autonomousResumeTriggerSchema.triggers`, widening a closed enum every shipped v1.1.x host validates via a strict `chatSchema.safeParse`. Any chat containing one becomes unreadable on v1.1.3 (`"Chat '<id>' could not be read from persisted state."`) once written by a v1.1.4 host.
- Moves wakeup triggers into an additive `wakeTriggers` key (a new defaulted array an old strict `z.object` parser silently strips) via a decode/encode codec on `autonomousResumeBlockSchema`. The domain shape (`triggers` only) is unchanged for every consumer other than the storage read/write funnels; legacy inline `kind: "wakeup"` entries from pre-fix builds are still accepted on read and re-encoded on next write.
- Adds a frozen `chatSchema` JSON-Schema surface tripwire (`chat-schema-surface-compat.test.ts` + `snapshot-chat-schema-surface.ts`) so any future persisted-schema change that isn't safely additive within persistence major 2 fails loudly instead of silently breaking older hosts again.

## Test plan
- [x] `content-blocks.test.ts`: legacy `kind:"wakeup"` inline acceptance, `wakeTriggers` merge ordering, frozen-old-schema degradation check, encode/decode round-trip, idempotency, `z.encode` on the union, registry-import/JSON-schema regression (15/15 passing)
- [x] New tripwire test verified to actually catch drift (temporarily widened an enum, confirmed a precise failing diff, reverted)
- [x] `hostStreamRpcRegistry` module-load-time validation (`z.toJSONSchema` both IO modes) confirmed to still succeed on the full `chatSchema`


hdkshingala:
@coderabbitai review

PR #255, opened 2026-07-08
Paginate worktree listing + worktree manager fixes (clickable tasks, PR chips, Landed tier)

tanveergill:
## Summary

Two related changesets on top of the existing worktree housekeeping feature:

**Pagination** (fixes a `worktree list --include-activity` timeout on hosts with many worktrees — the RPC frame timeout races the host's per-worktree probe pass):
- `worktree.listAllForHost@1.1` gains required `cursor`/`limit` request fields and a `nextCursor` response field, with two mutually-exclusive request modes (paged listing vs bounded per-path selection).
- CLI: `--cursor`/`--limit` flags, auto-page-to-completion by default, resume-on-failure envelope (`resumeCursor` + partial rows).
- gui-app: Settings ▸ Worktrees base list uses an infinite query with progressive rendering; task-delete candidates loop pages with probes restored (proof-based default-checking preserved).

**Worktree manager fixes** (live-tested feedback):
- Task chips are now clickable — open/focus the owning epic tab.
- PR chips row: one clickable chip per PR (superproject + every submodule branch), GitHub-vocabulary colors (merged=purple, open=green, closed=red), muted `<repo> · unmerged` chips for branch-bearing legs without a landed PR.
- The `merged` tier label renamed to **Landed** (enum id unchanged) — disambiguates the worktree-level cleanup verdict from a PR's own "Merged" state, since both now render as chips/pills side by side.
- `WorktreeClassification.facts` split structurally into `prFacts`/`nonPrFacts` so the GUI can render only non-PR facts (chips carry PR info now) while the CLI/skill reports keep the full set — no fact-string matching.

## Test plan

- [x] `bun run --cwd protocol test` / `compile`
- [x] `bun run --cwd clients/traycer-cli test` / `compile`
- [x] `bun run --cwd clients/gui-app test` / `compile`, react-doctor clean
- [x] `bun run --cwd clients/shared test`
- [x] Compat gate: compatible against all released baselines (v1.0 bridge preserved, probe-free)
- [x] Live-tested on the real dev fleet: pagination, clickable tasks, PR chips (merged/open/closed, superproject + submodule), Landed tier


tanveergill:
@coderabbitai review

tanveergill:
Addressed in b9841cc:

- **worktrees-enrichment.ts (cold-PR retry state):** the retry ledger effect is now keyed on \`[client, hostId, reachable]\`, so a host/client swap or reachability drop clears pending timers and the attempt map instead of carrying them into the next scope.
- **use-task-delete-worktree-candidates-query.ts (unbounded pagination):** the probe-page walk now fails closed — a hard \`TASK_DELETE_WORKTREE_MAX_PAGES\` cap plus cursor-cycle detection (repeated \`nextCursor\` throws) replace the unbounded \`for(;;)\`.
- **worktree-schemas.ts (limit):** tightened to \`z.number().int().positive().nullable()\`.

On the `worktrees-listing-query.ts` `refresh()` thread: that suggestion targets the wrong callback. `refresh` is the manual full-refresh button (a full `refetch()` is intended there). The partial-listing banner already retries via a separate `retryPartial: () => fetchNextPage()`, which is exactly the page-advancing primitive suggested — so no change is needed.

@coderabbitai review

tanveergill:
Follow-up in 802b161:

- **use-open-external-link-mutation.ts (namespace prefix):** renamed \`useOpenExternalLink\` → \`useRunnerOpenExternalLink\` to match the \`use<Namespace><Verb><Noun>\` convention (\`useRunnerCliLogin\`, etc.). All call sites updated; compile + lint green.
- **worktrees-listing-query.ts (`refresh()` re-fetch):** confirmed intentional — `refresh` is the manual full-refresh affordance; the partial-listing banner retries via the separate `retryPartial: () => fetchNextPage()`, which is already the page-advancing primitive suggested. No change.

@coderabbitai review

PR #266, opened 2026-07-09
fix(gui-app): stop macOS Ctrl chords matching mod keybindings

tanveergill:
## Problem

On macOS, Unix line-editing keys never reach a focused terminal: ⌃K opens the command palette, ⌃W closes the tab, ⌃D splits the pane, ⌃T/⌃N/⌃B/⌃Y/⌃L all fire app actions instead of their readline meanings. The same collapse also let ⌃1…⌃9 trigger the ⌘-digit actions and a bare Ctrl hold light the ⌘ leader badges.

## Root cause

The Keybindings settings screen publishes ⌘-only chords on macOS, but the event-matching path collapsed Meta/Ctrl into `mod` on **all** platforms (`parseChordFromEvent`, `modifierMaskFromEvent`, `cleanLeaderModifierFromEvent`). The `KeybindingProvider` reserves matched chords at window capture phase (`preventDefault` + `stopPropagation`), so a matched ⌃-chord died before xterm — or a native text input — ever saw it. The matcher accepted chords the UI never advertised.

## Fix

Make `mod` platform-aware everywhere it is derived from an event, via a shared `hasPlatformModKey(event)`: Meta-only on macOS, Meta-or-Ctrl elsewhere (non-mac behavior unchanged — Ctrl legitimately IS `mod` there). No terminal-focus special-casing: on macOS bare-Ctrl chords now never match `mod` bindings anywhere, matching the published contract.

- `chordMatchesEvent` prefers the ctrl-aware chord when it differs, so Ctrl-specific chords match exactly.
- `resolveReservedAction` no longer falls back to the lenient chord when the ctrl-aware chord differs, so a macOS Ctrl event can only match an explicit `ctrl+…` binding.
- Ctrl-specific bindings (voice input ⌃⇧M, model picker ⌃⌥M) and all ⌘ chords are unaffected; explicit user rebinds to Ctrl chords keep working (recorded ctrl-aware).

## Verification

- New/updated Vitest coverage: on macOS ⌃K/⌃W/⌃D/⌃1 no longer match mod bindings or digit actions, ⌘ equivalents still do, ⌃⇧M still matches its `ctrl+` chord; non-mac Ctrl still matches `mod`.
- `bun run compile`, `bun run test` (full suite), `bun run lint` all pass; react-doctor clean on changed files.


tanveergill:
Addressed in 3cc2ba2 — extracted `resolveMatchingChord(event)` in `chord.ts`; both `chordMatchesEvent` and the provider's `resolveReservedAction` now resolve the ctrl-aware precedence through that single helper. @coderabbitai

tanveergill:
Addressed in ec809ae — extracted the platform mock into `createPlatformMock()` in `src/__tests__/create-platform-mock.ts`, now shared by `chord.test.ts` and `keybinding-provider-visual.test.tsx`. @coderabbitai

PR #267, opened 2026-07-10
feat: fork a chat during Q&A (Cross Question / A/B Fork) + fork UI simplification

tanveergill:
## Summary

Adds the ability to **fork a chat while the assistant is asking a question** (a pending `AskUserQuestion` / interview), so the user can branch off without disturbing the original thread. Two flavors are offered on the pending-question card:

- **Cross Question** — forks on the chat's own working copy (binding verbatim: local stays local, an existing worktree is adopted). The carried question renders as inline reference and the composer is immediately free — for interrogating the assistant before answering.
- **A/B Fork** — forks into new worktrees off each folder's current branch, carrying uncommitted + staged changes; the carried question re-opens as an answerable card — for proceeding down an alternate path in parallel.

The per-message footer keeps a single plain **Fork conversation** button on completed messages (the two flavors are card-only).

## What's in this PR (submodule: gui-app + protocol)

- Relaxed fork gating so the affordance is available during a pending interview; `findPendingInterview` exposes the owning message id.
- Pending-interview card gains the two fork-flavor actions; per-message footer stays a single plain button.
- `forkSource.carriedInterviews` (`"pending" | "settled"`) drives whether the fork re-opens the carried question as answerable or settles it as reference.
- A/B fork rebases each folder to the chat's working-copy directory and pre-selects a new worktree carrying the working tree.
- Default fork title reflects the mode (Fork / Cross Question / A/B Fork).

The host-side changes (fork resolver, session-anchor handling for Claude/Codex/OpenCode) ride the paired `traycer-internal` PR.

## Testing

- `bun run compile` clean (gui-app).
- Vitest suites for the fork seed, interview card, and chat-tile pass; `react-doctor` clean on changed files.


tanveergill:
@coderabbitai thanks — quick responses:

1. **A/B pre-selection recovery** — intentional fail-safe. `canSubmitFork` gates an A/B fork on the staged worktree pre-selection because forking into a *new worktree* requires a resolved workspace; if the folder-summaries query never lands, the picker shows the folders as unresolved and there is genuinely nothing valid to fork into. Blocking submit (rather than silently falling back to the source binding, which would create the fork on the wrong working copy with no setup script) is the correct behavior; the user can cancel or retry. Verbatim (plain / cross-question) forks are not gated.

2. **buildForkWorkspaceSeed coverage** — this is a false positive; the `describe("buildForkWorkspaceSeed")` suite is still present (fork-workspace-seed.test.ts, line 154), alongside the new `buildAbForkWorkspaceSeed` tests.

No code changes needed for these.

tanveergill:
@coderabbitai resolve — the three comments are addressed (two are false-positive/intentional as noted above); resolving the threads.

PR #290, opened 2026-07-11
protocol: unit-test guard for released-line host→client growth

hdkshingala:
## Why

PR #285's first iteration added `devin` to the **released** v3.0 catalog schemas — and `protocol-compat` passed. The gate detected the divergence ("enum values added at a released version") but `surface-compat.ts` scores all additions as `advisory` on a feature-gating assumption that is correct for the ~180 in-place id-growth sites (chat frames, per-provider echoes — all reachable only through a negotiated catalog) and exactly wrong for the catalog responses themselves, which every released client strict-decodes unconditionally. Meanwhile the `protocol-surface.json` release metadata was published but consumed by nothing, and the unit-test layer froze method names only.

## What

Detection is now **fully automatic in both directions — no labels, no manual review** in the normal path:

1. **Direction-aware severity** (`surface-compat.ts`): enum-value/union-variant additions at a released negotiated version are `breaking` on host→client slots (`response`, `serverFrame`) and stay `advisory` on client→host slots. Failure text teaches the fix: freeze the shipped line, open a new major with downgrade bridges (amp `003d7586` / devin `407d110` template).
2. **Static catalog-gated policy** (`compat-exceptions.json`): 18 method-scoped, id-agnostic pattern entries encoding the practiced safe growth paths — future harnesses never touch this file. A load-time rule rejects any entry that would grandfather the host→client slots of `agent.gui.listHarnesses` / `agent.list` / `providers.list`; those can only ever be fixed by a new version line.
3. **The unit test** (`released-baseline-compat.test.ts`): the newest released surface (the same artifact releases publish as `protocol-surface.json`) is committed as a fixture, and the live registries are replayed against it through the shared oracle in plain `bun run test`. The ids-on-a-released-line mistake goes red locally with the teaching message; a correctly opened new major stays green with zero policy edits.
4. **Freshness automation** (`sync-released-baseline.yml`): on each host/cli/desktop release tag, regenerate the fixture and open a mechanical PR if it changed.
5. **Frozen catalog-line fixtures** (`frozen-catalog-lines.ts` + test): the historical frozen exports for the 3 catalog methods can no longer drift silently.

The tag-resolving `protocol-compat` CI workflow is untouched (byte-identical) and remains the zero-staleness authority — an in-tree fixture edit can never fool it.

## Validation

| Gate | Result |
| --- | --- |
| compile + protocol tests (incl. the new guard) + gui-app tests | protocol 712 passed · gui-app 4819 pass / 1 skip |
| Fixture regenerated by `snapshot-released-baseline.ts` vs the surface dumped from `cli-v1.1.6` | byte-identical |
| Regression: the incident tree (devin/pi ids on v3.0) vs the fixture | **exit 1 — exactly 3 blocking** (the catalog responses @3.0) |
| The corrected v4.0 tree (#285+#286) vs the fixture | **0 blocking**, zero policy edits needed |
| This tree vs all 10 released baselines (tag-resolved) | exit 0 |

## Notes

- This PR edits `compat-exceptions.json`, which is on the *pre-existing* tripwire's guarded list — that check will flag once here (the current gate protecting itself during its own upgrade). No new label ceremony is introduced anywhere.
- The fixture is ~3.5 MB of canonicalized JSON and changes only once per release via the sync workflow.
- Known papercut: `GITHUB_TOKEN`-authored sync PRs don't trigger CI runs (GitHub restriction); if branch protection requires checks, re-run via `workflow_dispatch` or switch the workflow to a bot token later.


hdkshingala:
@coderabbitai review

PR #328, opened 2026-07-13
fix: harden provider sign-in and add copy-link fallback

anur4ag:
Three coordinated improvements make provider profile sign-in reliable, recoverable, and easier to complete when opening the browser is inconvenient.

## 1. Transport: keep providers.awaitLogin alive for the OAuth window (f15aab4)

providers.awaitLogin is a long-poll: the host holds the response until the OAuth login child terminates, bounded by the host's 3-minute login timeout. The transport's 30-second per-request frame timeout previously misread that silence as a dead host and closed the socket mid-sign-in.

- Session.next() now accepts a per-call wait budget. Dial and openAck retain the default timeout, so unreachable hosts still fail fast; only the response wait is extended.
- IHostMessenger.requestWithResponseTimeout is threaded through retrying/auth-aware wrappers, HostClient, and the mock messenger; request() still delegates with the default.
- PROVIDERS_AWAIT_LOGIN_RESPONSE_BUDGET_MS is a 4-minute client-local budget derived from the host deadline plus probe slack. It is not a wire-contract change.
- Transport regressions cover late success, the unextended handshake, and expiry of the extended budget.

## 2. GUI: clear stale failure state on the next attempt (0c18b96)

The provider settings panel's failure banner used to survive a later successful attempt. AddProviderProfileDialog now clears the reported failure when a new attempt starts and again when an attempt completes. The full-journey regression covers failure, banner retry, clearing on restart, and success with the banner still absent.

## 3. GUI: add a copy-link fallback while sign-in is pending (b0f3977)

The pending step now offers a secondary Copy link action beside Open sign-in page. It copies the external authorization URL, shows transient Copied feedback, and reports clipboard failures through the standard reportable error toast. The in-flow cancellation action is visually destructive and displays Cancel while retaining the accessible name Cancel sign-in so existing role queries remain unambiguous.

This client change pairs with the host-side guarantee that loginUrl is the real external authorization URL rather than a localhost callback.

## Verification

- Full workspace compile: all 4 projects passed.
- Focused GUI journey suites: 38/38 tests passed.
- Existing transport and host-client regression suites passed on the earlier commits.
- React Doctor on changed GUI files: no issues.
- Pre-commit affected build, compile, lint, format, and DCO checks passed.


anur4ag:
CodeRabbit outside-diff suggestion triage: **no code change**. I verified the current `useHostMutation` and `useHostMutationWithResponseTimeout` implementations. This is a non-blocking maintainability suggestion rather than a correctness issue: the two thin typed wrappers deliberately make their distinct `HostClient.request` vs. `requestWithResponseTimeout` call paths explicit. Extracting a generic request callback solely for this one pair would add type/callback indirection without changing behavior or reducing a demonstrated drift risk. All current checks remain green.

PR #332, opened 2026-07-13
fix(notifications): clear host-backed notifications

tanveergill:
Fixes Clear all for host-backed notifications.

- Adds the optional host.notifications.clearAll contract and host-feed cleared frame.
- Clears the host-backed renderer store after a successful RPC and applies clear events from other windows.
- Preserves notifications newer than the latest loaded host notification.
- Adds UI, store, and protocol regression coverage.

Validated locally: focused GUI notification tests, protocol notification tests, GUI and protocol compile, and affected pre-commit compile/lint/format checks.


tanveergill:
@coderabbitai review

PR #339, opened 2026-07-14
feat(terminal): epic-less terminal sessions + landing terminal panel

hdkshingala:
Adds a multi-tab terminal panel to the start page, and the protocol/client
support that epic-less terminal sessions need.

Paired with the internal host PR that serves the independent scope; the two
must land together.

## Protocol

Terminal scoping becomes a discriminated union on the wire:

```ts
type TerminalScope = { kind: "epic"; epicId: string } | { kind: "independent" }
```

- `terminal.create@2.0` / `terminal.list@2.0` take `scope` in place of
  `epicId`, with v1<->v2 upgrade/downgrade bridges. The downgrade returning
  `ok: false` for an independent scope **is** the feature gate: a client-local,
  typed signal the GUI reads to hide the panel against an old host.
- `terminal.subscribe` deliberately stays **major 1**, shipped as `@1.4`.
  Streams have no cross-major bridges — `checkStreamCompatibility` fatals on any
  major mismatch and the manifest advertises only the highest major — so a
  `subscribe@2.0` would have broken terminal streaming for **every** released
  client, epic terminals included. The `@1.4` minor swaps a field inside the
  nested session-info sub-schema, which bends the additive-minor rule; it is
  wire-safe because the host emits frames per negotiated version and a `<=1.3`
  subscriber can never reach an independent session. Documented in-code; the
  additivity checker and `compat-exceptions.json` are untouched.

## Client

- `TerminalStreamClient` parses frames **per negotiated version** — the
  symmetric half of the host's per-version emit. Without this the host's
  scope-bearing `@1.4` frames are silently dropped by the `@1.3` schema and
  every terminal breaks, landing and epic alike.
- Terminal bootstrap takes an explicit `TerminalScope` instead of reaching for
  the ambient epic (which threw off-epic), and exposes `hostSessionExited`.

## GUI

Right-side terminal panel on the start page: multi-tab, resizable, collapse /
maximize, tabs styled like the epic strip.

- Sessions live host-side, so tabs survive reload and navigation; the panel
  reattaches with full scrollback.
- Ordered reconciliation on open / host change, strictly before auto-spawn:
  drain tombstones -> reattach -> adopt orphans -> classify leftovers.
- Close is **tombstone-first**: one atomic durable write removes the ref and
  records the pending kill *before* `terminal.kill` is dispatched, so a reload
  mid-kill can never re-adopt a closed shell.
- Tabs are host-bound for life; an unreachable host yields a dead tab rather
  than a silent migration.
- The panel hides entirely on hosts too old to serve the independent scope.
- `mod+j` toggles the panel, `mod+shift+j` opens a terminal — both bindable.

## Review follow-ups

Later commits address review findings. One is a **behaviour fix outside the
landing panel**, in the epic canvas:

- **Epic terminal tiles now close when they open onto an exited session.** The
  tile never consumed the bootstrap's `hostSessionExited`. Reopening a terminal
  whose PTY died inside the host's ~60s grace window left it on "Starting
  terminal session…", and once the grace lapsed — the session now *absent*
  rather than *exited* — the bootstrap quietly spawned a fresh shell under the
  same id, defeating the no-respawn gate by a minute. It now closes, as the
  landing panel drops its tab. Gated on never having attached, so a crashed
  terminal (deliberately kept on screen for its failure indicator, and reported
  `exited` by the host for the same 60s) is not ripped away.

Also: the epic-scoped visibility predicate is hoisted into
`isVisibleEpicTerminalSession` so the sidebar and palette opener stop restating
the rule that keeps a landing PTY out of an epic; both no-routable-host
fallbacks of the landing kill mutation are covered (they must fail, not reroute
a destructive RPC); and the `terminal.list` downgrade path uses functional
array methods.

## Verification

`bun run compile`, `bun run lint`, `bun run react-doctor` clean.
5,207 gui-app tests + 793 protocol tests pass, against `main` merged in.

Not yet driven end-to-end in the app.


hdkshingala:
@coderabbitai review

anur4ag (approved):

PR #347, opened 2026-07-14
Improve chat status indicators and tooltip accessibility

hdkshingala:
## Summary

- replace chat interview, approval, completion, failure, and read-only spinner variants with semantic message-square icons while preserving the in-progress spinner
- compose canvas title generation as the chat status default so running and notification semantics retain precedence
- consolidate workspace and owner previews on the shared Tooltip surface, moving interactive actions out of Tooltip content
- prevent overflowing Tooltip content from entering Chromium's sequential focus order and add accessible read-only status naming
- add cross-theme contrast guarantees for warning, success, destructive, copy, and inverse PR-pill presentation
- split the unreleased `host.notifications.indicatorState@1.0` prompt flag into approval/interview fields

## Compatibility

`indicatorState` is an unreleased development-only method and remains outside the released protocol floor. The shape can change in place on the flat 1.0 surface; no persisted schema migration, compatibility adapter, or release-specific workflow is required. The companion internal PR traycerai/traycer-internal#4397 updates the matching host aggregation and submodule pin.

## Validation

- GUI full suite: 641 files / 5,276 tests passed, 1 existing skip
- rebased focused GUI suite: 7 files / 86 tests passed
- protocol contract + released-floor: 15/15 passed
- GUI and protocol compile passed
- Desktop renderer production build passed
- affected Nx lint/format/build/compile pre-commit gate passed
- real Chrome 150 post-fix Tab walk skipped both visible and Radix-hidden Tooltip scroll roots


hdkshingala:
Internal host counterpart: traycerai/traycer-internal#4397. `indicatorState` is unreleased and development-only, so both PRs update the flat 1.0 shape together without a release-specific compatibility gate.

PR #362, opened 2026-07-15
fix(gui-app): anchor selection toolbar to tile scroll container

anur4ag:
## Summary

- thread the collab tile's live scroll-container element into `ArtifactToolbar`
- pass that element through memoized TipTap `BubbleMenu` options so the plugin rebinds its scroll listener when the ref arrives or changes
- stabilize `shouldShow` with `useCallback` to avoid metadata-only option updates on unrelated renders
- cover container-scroll repositioning, window → target A → target B listener migration, late editor mount, and unmount cleanup

## Root cause

TipTap's BubbleMenu plugin listens for scroll on `options.scrollTarget ?? window`. The tile's actual scroller is an inner `overflow-y-auto` element, and its scroll events do not bubble to `window`, so the toolbar never recomputed its position. In Chromium 149, a 300 px tile scroll moved the selection by -300 px while the menu moved 0 px.

The fix passes the tile-owned element directly to the plugin. TipTap's options-update path removes the prior listener and binds the new target, including the callback-ref/late-editor mount order.

## Verification

- `bun run compile`
- `bun run test` — 645 files passed, 1 skipped; 5,324 tests passed, 1 skipped
- `bun run lint`
- `bun run react-doctor` — no issues detected; internal lint analysis timed out after 300s, with standalone ESLint passing
- `pre-commit run --files <three changed files>` — passed, including affected build/compile/lint/format
- Chromium 149 E2E: exact 1:1 tracking in both directions; late-arrival rebind; zero positioning work while hidden; one debounced reposition per scroll burst

### DIFF
--- a/clients/gui-app/__tests__/editor-core/artifact-toolbar.test.tsx
+++ b/clients/gui-app/__tests__/editor-core/artifact-toolbar.test.tsx
@@ -1,7 +1,14 @@
-import { afterEach, describe, expect, it } from "vitest";
-import { render, screen, cleanup, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
 import { Editor } from "@tiptap/core";
-import { EditorContent, EditorContext } from "@tiptap/react";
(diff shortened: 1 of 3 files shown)


anur4ag:
Superseded by #363. The replacement starts from current main and includes the repository-required matching DCO sign-off without rewriting published history.

PR #479, opened 2026-07-19
fix(gui-app): quote final paragraphs cleanly

anur4ag:
## Summary

- allow last-paragraph triple-click quotes when Chromium extends the selection into elapsed-footer chrome with non-rendered SVG metadata
- strip visible data-quote-exclude tail text without flattening Chromium paragraph boundaries, including multi-option next-step groups
- drop terminal browser-synthesized blank lines before building blockquote paragraphs while preserving internal blanks and fenced code

## Root cause

Chromium can extend a last-paragraph triple-click range beyond the quotable root. The clamp guard treated invisible SVG title text as selected content and rejected the quote. Accepted clamps also need to preserve Selection.toString paragraph separators while removing rendered excluded chrome. Separately, trailing block-boundary newlines were being converted into empty paragraphs inside the blockquote.

## Verification

- pre-commit run --all-files
- bun run compile from clients/gui-app
- bun run test -- src/components/chat/quote/ — 69 tests passed
- ESLint with zero warnings on all four changed files
- Prettier check on all four changed files
- React Doctor changed-file scan — no issues
- DCO sign-off hook passed

## Related issue

N/A

## Checklist

- [x] Affected build, compile, lint, and format checks pass
- [x] Quote tests pass
- [x] Tests added and updated for the browser-specific regressions
- [x] Commits are signed off with DCO

### DIFF
--- a/clients/gui-app/src/components/chat/quote/__tests__/append-quote-to-draft.test.ts
+++ b/clients/gui-app/src/components/chat/quote/__tests__/append-quote-to-draft.test.ts
@@ -71,6 +71,50 @@ describe("buildQuoteBlockquote", () => {
     });
   });
 
+  it("drops a single trailing empty line (the browser's block-boundary blank after a triple-click)", () => {
+    const node = buildQuoteBlockquote({
+      text: "Third paragraph, the last one.\n",
+      fenceLanguage: null,
+    });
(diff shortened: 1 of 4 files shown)


anur4ag:
@coderabbitai approve

PR #482, opened 2026-07-19
fix(gui-app): hide closed app-wide stream client during rebuild

hdkshingala:
## Problem

The git-diff panel could get stuck on a permanent `CLIENT_CLOSED` skeleton instead of recovering like every other streaming RPC. The app-wide `WsStreamClient` is owned by React via `HostStreamProvider`. A replace-on-unmount race (`useCloseWsStreamClientOnReplace`'s deferred close firing across an effects-disconnect where React preserved the memoized value) could leave a **closed** client sitting in context. Every consumer of that client — git-diff, notifications, resources, worktree-changed, voice, migration — would then subscribe on the dead instance, receive an inert session, and surface a terminal `CLIENT_CLOSED` error with no path back to a live stream until a full window reload.

## Fix

Two-sided defense so a closed client is both hidden immediately and rebuilt reactively:

- **Consumer side** — `useWsStreamClient()` is now a `useSyncExternalStore` over the client's `onClosed` event whose snapshot returns `null` when `client.isClosed()`. A closed client is hidden the moment it closes (no rerender required), so consumers detach from the dead session and rebind once a replacement reaches context. The method-support readers (`useStreamMethodSupport` / `useStreamMethodSchemaVersion`) route through the same live-only hook.
- **Owner side** — `HostStreamProvider`'s existing liveness guard mints a fresh client via the rebuild nonce when one closes underneath it. The two halves compose cleanly: close → `null`, provider bumps the nonce once, the new open client changes context; `close()` is idempotent and the old guard subscription is cleaned up, so there is no rebuild loop.

Legitimate closes (host swap, sign-out, unmount) always coincide with a value change or teardown, so the guard's subscription is already gone before they fire; only the illegitimate replace-on-unmount race lands in it.

## Test

Adds a focused regression that exercises the exact `onClosed` subscribe branch the fix introduces:

1. Mounts with a **live** client and delivers a snapshot.
2. Closes the live client inside `act()` with **no parent rerender** — asserts the consumer detaches (its session closes) and **no `CLIENT_CLOSED`** surfaces. This only passes because `onClosed` notifies `useSyncExternalStore` on its own.
3. Swaps in a replacement client and asserts a **fresh git snapshot**.

Verified non-vacuous: temporarily neutering the `onClosed` subscription makes the new test fail precisely at the detach assertion.

## Verification

- Focused suite: 16/16 pass
- `bun run compile`: clean
- pre-commit (nx-affected build/compile/lint/format + DCO sign-off): pass

## Notes

Branch is based on the currently-pinned submodule commit (`424899d1`); it merges cleanly onto `main` (main did not touch these files). A follow-up worth considering: move the app-wide client out of React ownership into a module-owned holder keyed on `(hostId, userId)` to match the session-store model and delete the liveness guard / rebuild nonce / close-on-replace scaffolding outright.

### DIFF
--- a/clients/gui-app/src/hooks/git/__tests__/use-git-list-changed-files-subscription.test.tsx
+++ b/clients/gui-app/src/hooks/git/__tests__/use-git-list-changed-files-subscription.test.tsx
@@ -642,7 +642,7 @@ describe("useGitListChangedFilesSubscription", () => {
     expect(session.closed).toBe(true);
   });
 
-  it("a closed stream client surfaces an error via the inert session", async () => {
+  it("waits for a replacement client and recovers without surfacing CLIENT_CLOSED", async () => {
     const closedClient = new WsStreamClient<HostStreamRpcRegistry>({
       registry: hostStreamRpcRegistry,
       endpoint: () => null,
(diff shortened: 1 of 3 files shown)


hdkshingala:
Addressed the CodeRabbit maintainability note (outside-diff comment on the git subscription test) in 740c349e: extracted `makeSwappableStreamWrapper` and `swapToReplacementClientAndAssertHeadSha`, and routed both replacement-recovery tests through them. The unique detach-without-rerender assertion stays inline. Coverage unchanged — 16/16 pass, and the detach test remains non-vacuous (it fails at `expect(liveSession.closed).toBe(true)` when the `onClosed` subscription is neutered).

PR #487, opened 2026-07-20
fix(desktop,cli): split SMAppService agent label from the CLI's launchd label

hdkshingala:
## Problem

1.1.7-rc.2 started shipping an in-bundle SMAppService LaunchAgent plist in release builds (#389), routing release-install `host ensure` through the desktop's SMAppService register cycle for the first time. On any machine upgraded from a CLI-registered install (<= 1.1.6), macOS BTM matches that registration to the existing **legacy** record under the same label (`ai.traycer.host`) instead of creating a fresh app-scoped one — and that legacy record survives file deletion, bootout, and BTM sweeps. Registration lands `not-registered` permanently; `host ensure` fail-fasts and the desktop gets stuck on an error page that Retry cannot recover from.

Confirmed on `backgroundtaskmanagementd`:
```
registerLaunchItem: found existing item: … type=legacy agent,
  disposition=[disabled, allowed, not notified], identifier=8.ai.traycer.host
registerLaunchItem: updated item with same LWCR: … disposition=[disabled, …]
effectiveItemDisposition: record not found: appURL=/Applications/Traycer.app, type=app
```

## Fix

Split the two registration mechanisms onto separate labels. The CLI keeps `ai.traycer.host[.env]` for raw launchctl installs; the desktop's SMAppService agent registers under the derived, collision-free `<cli-label>.agent`. This also permanently immunizes the CLI-install → desktop-install transition, which would otherwise recreate this exact collision for every future headless-CLI user who adopts the app.

- `host-paths.ts` / `service/label.ts` — single-sourced `.agent` derivation on each side (desktop and CLI can't import each other; the packaging injector and the internal installer carry their own copies, all pinned to the same rule).
- `host-login-item.ts` — the register cycle, once committed (past both the removed-by-user and busy-revalidation guards), now also retires the legacy label: removes the stale CLI manifest, boots out the legacy job, and unregisters the old SMAppService serviceName — before registering the agent plist. Legacy cleanup is coupled to the committed cycle so a deferred/busy cycle never touches a still-running legacy host.
- `inject-host-launch-agent.cjs` — packages both the registered agent plist and an inert copy of the old-label plist, so the transition unregister above can resolve a pre-split app-scoped BTM record and drop it cleanly (no ghost Login Items row) on already-SMAppService-registered machines.
- `service/platforms/macos.ts` — `installService` refuses to bootstrap beside a desktop-owned agent (prevents a dual-host setup); `statusService` reports `externally-managed` for agent-owned machines; `uninstallService` and `stop`/`start`/`restart` are agent-label aware so manual CLI commands route to the Desktop app instead of hanging or failing on a launchd job that no longer exists under the old label.
- `doctor/engine.ts` — externally-managed copy no longer implies the bare CLI label is the one actually registered.

## Verification

- `bun run compile` — clean across all 5 projects.
- `traycer-cli`: 523/523 passing (+1 skipped).
- `desktop`: 639/639 passing.
- Prettier clean.
- **Live-validated end to end** on a real machine's poisoned BTM state: re-manufactured the exact field collision (legacy plist + loaded job + dangling BTM record), installed the fixed build, and confirmed the cycle runs in order (legacy manifest removed → legacy job booted out → old serviceName unregistered → agent registered) and lands `enabled` in ~90ms — versus permanently `not-registered` before the fix — with the host provisioning in under 5s and the production host on that machine unaffected throughout.
- Independent fresh-agent review (`traycer-review`, Claude Fable xhigh) — verdict: safe to land. All findings folded in above (agent-label awareness for stop/start/restart, a CLI↔desktop label-derivation parity test, the doctor copy fix). One accepted, no-change-requested finding remains: a rare, self-healing transient dual-host window if the old-serviceName unregister fails while the agent register succeeds mid-transition.

## Follow-up (not in this PR)

The inert old-label plist should be removed once the fleet has cycled through a few post-split releases — tracked separately, not blocking this fix.


hdkshingala:
Addressed both threads in 1878b0e0.

**Test coverage gap (macos.test.ts:1116)**: applied the suggested restructuring — the "proceeds normally when not-loaded" test now loops over `stop`/`start`/`restart` instead of only `start`.

**Uninstall bootout loop (macos.ts:513)**: fixed the real bug (a hard failure on the agent-label target skipped the CLI-label bootout entirely) by collecting failures per target and attempting both before reporting. One deliberate deviation from the suggested diff: `rm(serviceManifestPath(...))` stays gated on both bootout attempts being clean (success or benign not-loaded), not unconditional — `statusService` treats a missing manifest as `not-installed`, so removing it after a genuinely failed bootout would misreport a still-loaded job as gone. This is also the pre-existing invariant the "surfaces a real bootout failure and preserves the service manifest" test already pinned; kept it intact and added a new regression test (`attempts the CLI-label bootout even when the agent-label bootout fails hard...`) that pins both properties together.

hdkshingala:
Fixed in c444c7d1 — applied the suggested strengthening verbatim (pinning each bootout call's target string and order, not just the command-name count).

PR #508, opened 2026-07-20
fix(shared,gui-app): keep client resilient across renderer suspension

tanveergill:
## Problem

Resume-after-suspension (macOS App Nap / an occluded overnight window) freezes the renderer's JS. On wake the first stream re-dial carried an **expired bearer** and burned a round-trip on a certain `UNAUTHORIZED`, and every such wake surfaced a **"sign in" toast** even though recovery succeeded seconds later.

## Changes

- **`ws-stream-client`** — before dialing, read the bearer's `exp` locally (advisory, unverified) and, when already expired, run the existing single-flight `revalidateThenReconnect` first so the dial carries the rotated token. N reconnecting sessions coalesce into one refresh. The reactive `UNAUTHORIZED` path stays the authority for everything the local read can't see (revocation, clock skew, config mismatch); an undecodable token falls through to a normal dial.
- **`host-error-toast`** — suppress the "Please sign in again." toast for a *recoverable* `UNAUTHORIZED` while the app session is still signed in. The authoritative sign-out toast still fires via `AuthSessionExpiredToastBridge` after a real `signOut()`; the host-side JWKS `retryable` variant keeps its distinct copy.

Pairs with the host-side protocol-ping change (internal repo) so a suspended-but-alive renderer isn't reaped; each degrades gracefully without the other.

## Verification

`@traycer-clients/shared` + `@traycer-clients/gui-app` compile and lint clean; shared transport tests pass; the three gui-app tests asserting the "Please sign in again." copy pass (18/18).


tanveergill:
@coderabbitai full review

Issue #541, opened 2026-07-21
Windows host drops user PATH, making installed developer tools unavailable to agents


nsxdavid:
### Component

Desktop app

### Version

1.1.7

### OS / platform

Windows 11 Pro, x64, build 26200

### What happened?

Codex chats started from Traycer receive a minimal `PATH` instead of the normal Windows user and machine `PATH`.

This makes installed developer tools unavailable to the agent. In the affected session, none of these commands can be resolved:

- `git`, `gh`
- `node`, `npm`, `npx`, `pnpm`, `bun`
- `python`, `python3`, `uv`
- `cargo`, `rustc`
- `docker`, `kubectl`
- `rg`, `code`

The first visible symptom was a ClickUp MCP server configured through `npx`. It works in Codex CLI but is missing from Codex chats started from Traycer because Codex cannot start `npx`.

I expected a Traycer Codex chat to have access to the same installed tools and configured MCP servers as Codex CLI.

### Steps to reproduce

1. Install Node.js and confirm `npx` is available in a normal terminal.
2. Configure an MCP server in `~/.codex/config.toml` using `command = "npx"`.
3. Confirm the MCP server works in Codex CLI.
4. Start a new Codex chat in Traycer.
5. Check whether the MCP tools and other installed developer commands are available.

The MCP tools are missing, and commands available in the normal terminal cannot be resolved.

Adding the normal Node.js and npm directories to `PATH` before launching the same Codex runtime makes the MCP server load successfully.

### Logs / output

```shell
PATH inside the Traycer Codex session:

C:\Program Files\PowerShell\7
C:\Users\simu\.codex\tmp\arg0\...
C:\Users\simu\.traycer\cli\bin

Missing directories include:

C:\Program Files\nodejs
C:\Users\simu\AppData\Roaming\npm
```

The current workaround is to add the missing directories to Traycer's environment settings and restart the host.

PR #673, opened 2026-07-25
fix(windows): preserve env override casing and explain shell behavior

AmiteshwarRandhawa:
## What changed

- apply environment overrides case-insensitively on Windows while preserving the inherited key's original casing
- add a shared Windows shell-family classifier for PowerShell, Git Bash, WSL, and other shells
- explain in Shell settings how the selected Windows shell affects spawned agent environments
- document the behavior and add protocol/UI regression coverage

## Why

The host-side fix for #541 reconstructs a normal Windows environment, where PATH is commonly keyed as `Path`. A user override keyed as `PATH` must update that value in place instead of creating a second colliding entry. The settings caption also makes the PowerShell, Git Bash, WSL, and registry-only behavior explicit to users.

## Impact

Windows overrides now follow the platform's case-insensitive environment semantics without changing POSIX behavior. Windows users can also see whether agents receive shell-profile additions or only the registry-merged environment.

Companion host PR: https://github.com/traycerai/traycer-internal/pull/4625

Part of #541

## Validation

- protocol tests: 18 passed
- Shell settings caption tests: 5 passed
- `bun run compile` in `protocol/` and `clients/gui-app/`
- protocol and GUI workspace lint
- React Doctor: no issues
- affected-workspace pre-commit build, compile, lint, and format checks

### DIFF
--- a/clients/gui-app/src/components/settings/SETTINGS.md
+++ b/clients/gui-app/src/components/settings/SETTINGS.md
@@ -420,7 +420,18 @@ dialog.tsx` / `notification-hook-draft.ts`, unchanged by this pass).
     when the selected program is a login shell, and a quiet _Restore default
     flags_ action shown only while the visible flags deviate from the family
     default - reverting the SELECTED shell via
-    `useRunnerTraycerShellRevertArgsMutation`); and a **`"Host environment ·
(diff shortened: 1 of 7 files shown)


AmiteshwarRandhawa:
@codex how do I resolve all your comments?

PR #677, opened 2026-07-25
fix(gui-app,protocol): keep a revalidating provider's models on screen

tanveergill:
## Problem

Opening the model picker sometimes showed a blank list that filled back in a few seconds later.

The host's per-adapter availability cache lapses every 30 s, while the GUI re-lists harnesses on a 15-minute poll lane. So nearly every catalog refresh found the entry lapsed and got back `availabilityPending: true` for harnesses the host was merely **re-probing**. The picker read that as unavailable and retired a catalog it already held:

- `useGuiHarnessCatalog` only builds model queries for available harnesses, so the entry's `models` fell back to `[]`
- the picker's `rows` filters on `harness.available`
- `ModelRowsState` shows "provider unavailable" / the API-key CTA before it ever looks at cached rows
- every rail entry went spinner + `aria-disabled`, and ⌘-digit switching was refused

800 ms–5 s later the fast-poll settled and everything reappeared. Nothing was ever evicted (`gcTime: Infinity`) — the list was hidden.

## Change

`availabilityPending` now only means "a probe is running". `available` carries the last settled verdict, so a lapsed-cache revalidate reports `available: true` (see the paired host-side stale-while-revalidate change).

`harnessAvailabilityUnsettled(harness)` names the one state that must stay inert — probing with **no** verdict ever settled — and the rail's spinner, disabled state, aria labels, tooltip and leader-digit dispatch all key off it instead of `availabilityPending` alone.

A *settled* unavailable verdict still retires the list and shows the provider's setup CTA. Serving models for a provider that is genuinely gone would trade a visible flicker for a failing send.

Protocol edits are **comments only** — no schema change, no version change.

## Verification

- new picker regression test: a revalidating provider keeps its rows on screen, stays clickable, and still commits on switch (verified red without the fix)
- `harness-model-picker` 55/55; picker + rail + catalog suites 92/92
- `bun run compile` clean across the graph; eslint clean on changed files


tanveergill:
@coderabbitai review

Base merges dismissed the earlier review (`dismiss_stale_reviews_on_push` + `require_last_push_approval` are on for `main`). Requesting a review of the current head `b3e57745` — the only substantive change is the 6-file picker/protocol diff; everything else is the merge from `main`.

tanveergill:
@coderabbitai review

Updated to `main` again (`7e868622`), which dismissed the prior approval. Substantive diff is unchanged — the 6 picker/protocol files; everything else is the merge from `main`.

tanveergill:
@coderabbitai review

Re-synced with `main` (`182775bd`); the push dismissed the prior approval. Substantive diff is unchanged - the 6 picker/protocol files.

PR #695, opened 2026-07-25
feat(gui-app): show Oh My Pi's PATH-only install guidance

### LINKED ISSUE #696: Oh My Pi has no PATH-only install guidance in provider settings
### Problem / motivation

Oh My Pi (`omp`) landed in #675 with the full protocol line and renderer surface, but not the
**PATH-only not-found state**. omp ships without a bundled binary, exactly like Hermes — yet only
Hermes tells the user so.

Today an omp user who has not installed the CLI opens **Settings → Providers → Oh My Pi** and sees a
bare `Path` / `Version` table: no rows, no explanation, and no link to installation instructions.
Hermes in the identical situation gets a short paragraph plus a deep link to its install guide.

The cause is that the guidance is gated on a single hard-coded provider id in
`clients/gui-app/src/components/settings/panels/provider-cli-candidates-section.tsx`:

```ts
const isEmptyHermesState =
  providerId === "hermes" && cliConfig.candidates.length === 0;
```

Any future PATH-only provider inherits the same gap by default.

### Proposed solution

Replace the single-provider branch with a lookup keyed by provider id, so every PATH-only provider
shares one code path:

```ts
const PATH_ONLY_INSTALL_GUIDES: Partial<Record<ProviderId, PathOnlyInstallGuide>> = {
  hermes: { name: "Hermes",   url: "https://hermes-agent.nousresearch.com/..." },
  omp:    { name: "Oh My Pi", url: "https://omp.sh/#install" },
};
```

The next PATH-only provider then becomes a table row rather than another branch. The shape mirrors
`API_KEY_DASHBOARD_URL` in the sibling `provider-api-key-section.tsx`.

Hermes rendering stays byte-identical — same gate, same copy, same URL — so its existing PATH-only
test covers the refactor unchanged.

Implemented in PR #695.

### Alternatives considered

- **Add a second `isEmptyOmpState` branch beside the Hermes one.** Rejected: duplicates the JSX block
  and leaves the same gap for the provider after omp.
- **Derive the label from `PROVIDER_DISPLAY_NAMES`.** Rejected: that map holds product names
  (`"Hermes Agent"`), which read wrong in "<name> must be installed on this machine", and adopting it
  would change Hermes' already-shipped copy.
- **Link `https://omp.sh` (site root).** Rejected in favour of `https://omp.sh/#install`, which lands
  on the install section and matches the semantics of the Hermes entry's deep link.

### Component

Desktop app

metaphorics:
## Summary

`omp` landed in #675 with the full protocol line and renderer surface, but not the
**PATH-only not-found state**. Oh My Pi ships without a bundled binary, exactly like
Hermes — yet only Hermes tells the user so. Today an omp user who hasn't installed the
CLI opens **Settings > Providers > Oh My Pi** and gets a bare `Path`/`Version` table:
no rows, no explanation, nowhere to go.

Hermes' arm was a hard-coded `providerId === "hermes"` branch, so bolting omp on beside
it would have meant a second copy of the same JSX. Instead it becomes a lookup:

```ts
const PATH_ONLY_INSTALL_GUIDES: Partial<Record<ProviderId, PathOnlyInstallGuide>> = {
  hermes: { name: "Hermes", url: "https://hermes-agent.nousresearch.com/..." },
  omp:    { name: "Oh My Pi", url: "https://omp.sh/#install" },
};
```

Both providers share one code path, and the next PATH-only provider is a table row
instead of another branch. The shape mirrors `API_KEY_DASHBOARD_URL` in the sibling
`provider-api-key-section.tsx`.

Three details worth calling out:

- **Keyed by provider id, never by display name.** `"Oh My Pi"` contains the separate
  `"Pi"` provider's name as a substring — the same collision #675 flagged.
- **`omp.sh/#install`, not the site root.** omp.sh is a single landing page whose install
  section is `<section id="install">`; the anchor lands where the "installation guide"
  label promises, matching the Hermes entry's deep link. Verified in a browser.
- **`name` is deliberately not `PROVIDER_DISPLAY_NAMES`.** That map holds product names
  (`"Hermes Agent"`), which read wrong in "<name> must be installed on this machine."
  Keeping the short form also leaves Hermes' shipped copy byte-identical.

**Hermes rendering is unchanged** — same gate, same copy, same URL. Its existing
PATH-only test is untouched and is what covers the refactor.

Not addressed here: the empty state still doesn't consult `availabilityPending`, so it
can flash during a probe. That's pre-existing Hermes behaviour and out of scope for this
change.

## Related issue

Closes #696

Follow-up to #675 (`omp` harness).

## Checklist

- [x] Pre-commit hooks pass (`pre-commit run --all-files`)

<sub>Verification run: `pre-commit run --files <the two touched files>` → all hooks
Passed, including `build · compile · lint · format (nx affected)`. Full gui-app suite:
**7097 passed / 1 skipped / 0 failed** across 765 files. Rebased on `main` @ `c531352b`.</sub>

### DIFF
--- a/clients/gui-app/src/components/settings/panels/__tests__/providers-settings-panel.test.tsx
+++ b/clients/gui-app/src/components/settings/panels/__tests__/providers-settings-panel.test.tsx
@@ -996,6 +996,47 @@ describe("<ProvidersSettingsPanel />", () => {
     ).toBeNull();
   });
 
+  it("shows Oh My Pi's PATH-only not-found guidance without a bundled row", () => {
+    providerMocks.listResult.data = {
+      providers: [
+        providerState({
+          providerId: "omp",
+          selected: { kind: "path" },
+          candidates: [],
+          envOverrides: [],
+        }),
+      ],
+    };
+
+    render(
+      <RunnerHostContext.Provider value={createRunnerHost()}>
+        <TooltipProvider>
+          <ProvidersSettingsPanel />
+        </TooltipProvider>
+      </RunnerHostContext.Provider>,
+    );
+
+    expect(
+      screen.getByText(
+        "Oh My Pi must be installed on this machine. It ships without a bundled binary.",
+      ),
+    ).toBeDefined();
+    const guide = screen.getByRole("link", {
+      name: "Oh My Pi installation guide",
+    });
+    expect(guide.getAttribute("href")).toBe("https://omp.sh/#install");
+    fireEvent.click(guide);
+    expect(providerMocks.openExternalLink).toHaveBeenCalledWith(
+      "https://omp.sh/#install",
+    );
+    expect(
+      screen.getByRole("button", { name: "Add custom path" }),
+    ).toBeDefined();
+    expect(
+      screen.queryByRole("radio", { name: "Select bundled binary" }),
+    ).toBeNull();
+  });
+
   it("orders the provider rail by the default provider order", () => {
     providerMocks.listResult.data = {
       providers: [
--- a/clients/gui-app/src/components/settings/panels/provider-cli-candidates-section.tsx
+++ b/clients/gui-app/src/components/settings/panels/provider-cli-candidates-section.tsx
@@ -24,8 +24,33 @@ import { cn } from "@/lib/utils";
 // table past the panel width.
 const TABLE_GRID =
   "grid grid-cols-[2.25rem_minmax(0,1fr)_minmax(5.5rem,auto)_2.25rem] items-center";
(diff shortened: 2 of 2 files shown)


hdkshingala:
Thank you for looking into this, @metaphorics!

For Hermes, we couldn't bundle the CLI due to packaging limitations, but with OMP, we'll ship the binaries bundled with our host.

Image

You are seeing the issue because the OMP is still in the early testing phase, so the host has not been released yet. Once the host is released, you will see the bundled as well as the "Add custom path" option.

PR #701, opened 2026-07-25
fix(gui-app): restore composer focus after pane jump

tanveergill:


tanveergill (commented):
Addressed the requested test-query cleanup in 4077ff6a; focused suites and the full affected pre-commit gate pass.

PR #772, opened 2026-07-28
fix(auth): stop device-flow poll interval from ratcheting past approval

hdkshingala:
## Summary
- Users saw a 30-40s delay between approving device-code sign-in in the browser and the client continuing. Once paired with the companion server fix (internal `authn-v3` PR), the client no longer holds onto a widened poll interval past the violation that caused it: `DevicePollSchedule` now decays back to the server-issued base interval after any accepted `authorization-pending` poll, in both the desktop Electron loop and the CLI loop.
- The GUI sign-in panel no longer shows "Waiting for approval" through token validation/persistence after the poll already returned `authorized` — it flips to "Approved - finishing sign-in" the moment the token is in hand, via a new `phase: "waiting-approval" | "finalizing"` on `DeviceFlowProgress`.
- Cold-reviewed (independent agent pass): flagged the interval decay as a deliberate deviation from RFC 8628 §3.5 ("all subsequent requests" should stay paced) — kept intentionally and documented in the `resetPollInterval` doc comment, since this client only talks to Traycer's own authn (whose gate is itself fixed, non-ratcheting) and undecayed widening would otherwise pin a padded interval on the CLI (no browser-return nudge) for an attempt's remaining life after one spurious `slow_down`.
- Ships independently of the paired `authn-v3` server fix in either order: an old server still benefits from the decay (bounds how long a real `slow_down` interval lasts), and a new server's terminal-before-throttle change alone already removes the bulk of the reported delay.

## Test plan
- [x] `clients/shared/auth/__tests__/device-auth.test.ts` — `resetPollInterval` unit test (26/26 passing)
- [x] `clients/desktop/src/electron-main/auth/__tests__/device-flow-controller.test.ts` — end-to-end fake-timer loop test pinning the decay (8/8 passing)
- [x] `clients/traycer-cli/src/auth/__tests__/login-flow.test.ts` — unaffected, still green (4/4 passing)
- [x] `clients/gui-app/src/lib/auth/__tests__/auth-service.test.ts` — waiting → finalizing → cleared phase transition (82/82 passing across gui-app suite)
- [x] `clients/gui-app/src/components/layout/__tests__/sign-in-button.test.tsx` — updated fixture literal, still green (7/7 passing)

### DIFF
--- a/clients/desktop/src/electron-main/auth/__tests__/device-flow-controller.test.ts
+++ b/clients/desktop/src/electron-main/auth/__tests__/device-flow-controller.test.ts
@@ -167,6 +167,53 @@ describe("DeviceFlowController", () => {
     ]);
   });
 
+  it("decays the poll interval back to base after a compliant pending poll", async () => {
+    let tokenCalls = 0;
+    restoreFetch = installFetch((url) => {
+      if (url.endsWith("/device/authorize")) {
+        return authorizeOk();
+      }
+      if (url.endsWith("/device/token")) {
+        tokenCalls += 1;
+        if (tokenCalls === 1) {
+          // slow_down widens the 1s base interval to 6s (+5s, no Retry-After).
+          return new Response(JSON.stringify({ error: "slow_down" }), {
+            status: 429,
+            headers: { "Content-Type": "application/json" },
+          });
+        }
+        return tokenCalls === 2
+          ? new Response(null, { status: 428 }) // authorization-pending
+          : tokenAuthorized();
+      }
+      return new Response(null, { status: 500 });
+    });
+
+    const results: DeviceFlowResultPayload[] = [];
+    const controller = new DeviceFlowController(AUTHN, RETURN_SCHEME);
+    const outcome = await controller.start({
+      onResult: (_attemptId, result) => results.push(result),
+    });
+    expect(outcome.ok).toBe(true);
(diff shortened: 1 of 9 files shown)


hdkshingala:
Verified and fixed in 9d88b351.

You're right: `authResolvedDuringStart` was set before the first await on every other terminal outcome of an interactive device-flow attempt (launch failure, denied/expired, timeout) but was missing from the `authorized` branch. Confirmed the race is real by temporarily reverting the fix and running a new regression test that:

1. Seeds a stale, independently-valid stored session ("stale-user").
2. Calls `signIn()`, then invokes `start()` again *after* `identityGeneration` has already bumped — so the overlapping `start()` shares the exact generation the live attempt is using, and the generation fence alone cannot distinguish them.
3. Authorizes the live attempt with a fresh token/identity and confirms it lands.
4. Releases the stale `start()`'s deferred validation.

Without the fix, the stale rehydration clobbers the freshly authorized session (`AssertionError: expected 'stale-token' to be 'fresh-token'`). With the fix (setting `authResolvedDuringStart` first, matching the existing convention), the test passes and the authorized identity survives.

`bunx vitest run src/lib/auth/__tests__/auth-service.test.ts src/components/layout/__tests__/sign-in-button.test.tsx` → 83/83 green.

Issue #33, opened 2026-01-26
Ticket creation fails when system goes to sleep/lock - no progress recovery


ameet-personal:
## Describe the Issue

Provide a clear and concise description of the bug.

## Type of Issue

- [x] Performance
- [ ] Security
- [x] Usability
- [ ] Other

## Language

- [ ] JavaScript
- [x] TypeScript
- [ ] Python
- [ ] Golang

## Version Information

- Extension version: (e.g., "v0.1.0")
- VS Code version: (e.g., "1.89.1")
- OS version: (e.g., "Ubuntu 20.04")

## Additional Context

Add any other context or screenshots about the issue here.

## Environment
- Traycer version: 2.14.8
- Cursor version: 2.4.21
- OS: macOS arm64 25.2.0

## Description
When creating multiple tickets (5 for Phase 1), if the system goes to sleep or the screen locks during the process, Traycer fails and loses progress. There's no recovery mechanism.

## Steps to Reproduce
1. Start creating multiple tickets (long-running operation)
2. Let the system go to sleep or lock screen
3. Wake/unlock the system
4. Traycer has failed, progress lost

## Expected Behavior
- Either prevent system sleep during long operations
- Or save progress periodically so it can resume
- Or gracefully recover on system wake

## Actual Behavior
Process fails completely, must restart from scratch.

## Impact
Combined with the high CPU usage (issue #X), this makes ticket creation extremely fragile. Users must babysit the process.

ameet-rajababa:
Related to #23 which requested session preservation. This issue shows why that feature is needed - long-running operations like ticket creation can fail and lose all progress if the system sleeps.

Issue #32, opened 2026-01-26
Extension causes high CPU load (100%) when creating tickets - regression of #4, #8, #13


ameet-personal:
## Environment
- **Traycer version:** 2.14.8
- **Cursor version:** 2.4.21
- **VSCode version:** 1.105.1
- **OS:** macOS arm64 (Apple Silicon) 25.2.0
- **Plan:** Pro+

## Description
When using Traycer to create tickets from a validated spec, the Cursor Helper process spikes to 100% CPU and stays there indefinitely. This makes the system extremely slow and unresponsive. Ticket creation that should take seconds takes 10+ minutes.

## Steps to Reproduce
1. Open Cursor with Traycer extension installed
2. Create a spec with multiple phases (I had 5 tickets for Phase 1)
3. Ask Traycer to generate/create tickets
4. Observe Activity Monitor - "Cursor Helper" process at 98-100% CPU

## Expected Behavior
Ticket creation should complete in reasonable time without pegging CPU.

## Actual Behavior
- CPU stays at 100% for extended periods (10+ minutes)
- System becomes sluggish
- Fan spins up
- Ticket creation eventually completes but takes far too long

## Notes
- This is NOT a rate limiting issue - I have 8/15 artifacts available
- This appears to be a regression of issues #4, #8, and #13 which reported the same "high CPU load" problem
- Restarting Cursor temporarily fixes it, but CPU spikes again when ticket creation resumes

PR #795, opened 2026-07-29
fix(desktop): name the macOS Login Items entry after the product

hdkshingala:
## The defect

System Settings → Login Items shows the bundled host agent as **`traycer-host-start`** — a filename leaking into a product surface.

BTM records its `Name` from the **verbatim basename of the plist's `BundleProgram`**, and is not bundle-aware at all.

## Probe evidence

The backlog's hypothesis was *"point `BundleProgram` at the helper's `CFBundleExecutable` and BTM will show `Traycer Host`"*. That is **wrong**, and this PR is a different shape as a result.

Probed live on macOS 26.5: a throwaway app registered four variants through the real `SMAppService.agent(plistName:)`, each with deliberately distinct name fields, read back with `sfltool dumpbtm`.

| `BundleProgram` basename | helper bundle's name fields | BTM `Name` |
|---|---|---|
| `probe-host-start` (control) | display=`ProbeDisplayName`, name=`ProbeBundleName`, dir=`Probe Host.app` | `probe-host-start` |
| `probeexec` — the helper's own `CFBundleExecutable` | same | `probeexec` |
| `Probe Launcher Named` | same | `Probe Launcher Named` |
| `Probe Host` + `RunAtLoad` | same | `Probe Host` (exec'd, exit 0) |

`CFBundleDisplayName`, `CFBundleName` and the `.app` directory name were **all ignored** — including in the variant pointing at `CFBundleExecutable`, which would have shipped `Name: traycer`. **Renaming the launcher file is the only lever** — the same conclusion the CLI path reached for `/bin/sh` in #790.

The fourth arm confirms a space-containing basename both registers *and* gets exec'd by launchd with `$0` intact.

## The change

`HOST_START_COMPAT_LAUNCHER` becomes `` `${PRODUCT_NAME} Host` `` — derived, not hardcoded, so a product rename can't strand a stale name in Login Items.

**Not unified with the CLI's `HOST_START_LAUNCHER_BASENAME`, deliberately.** That constant is what `isHostStartInvocation` matches for the CLI's `~/.traycer/service/<label-id>/…` launcher form, so renaming *it* is a field-compatibility and eviction-attestation change. This bundled agent never matched that arm — its launcher's parent directory is `MacOS`, not the label id, so it attests through its label signal. That's precisely what makes this basename free to choose for the human reading it.

## Tests

A dedicated test asserts the basename, that the file exists and is executable at that path, and that `ProgramArguments[0]` and `BundleProgram` name the same file.

Each assertion was mutation-probed. That caught a real problem: the first version used `toContain("Traycer Host")`, which is **vacuous** — the helper directory is `Traycer Host.app`, so it passes for any launcher name. Reverting the constant left it green. The assertions now pin the full path and go red on revert.

## Lockstep

`scripts/desktop-install-cloud.js` in the internal repo is the twin and carries the same change; its PR is linked in a comment below once opened.

### DIFF
--- a/clients/desktop/scripts/prepack/__tests__/inject-host-launch-agent.test.ts
+++ b/clients/desktop/scripts/prepack/__tests__/inject-host-launch-agent.test.ts
@@ -22,7 +22,9 @@
 import { afterEach, describe, expect, it } from "vitest";
 import { createRequire } from "node:module";
 import {
+  accessSync,
   chmodSync,
+  constants,
   cpSync,
   existsSync,
   mkdirSync,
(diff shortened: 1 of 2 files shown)


hdkshingala:
Internal lockstep twin: traycerai/traycer-internal#4685 (same change to `scripts/desktop-install-cloud.js`, the cloud-desktop installer's copy of this launcher-writing logic).

hdkshingala:
## Confirmed against the real signed release (not just the ad-hoc probe)

The probe app in the description was **ad-hoc signed**, which left one arm open: does a properly Developer-ID-signed, notarized build get different treatment from BTM? Read back from the installed `/Applications/Traycer.app` on macOS 26.5 with `sfltool dumpbtm`:

| BTM record | `Name` | `Developer Name` | Type |
|---|---|---|---|
| #32 — parent app | `Traycer` | `Traycer AI, Inc.` (team `7YVZ56DZ74`) | app |
| #33 — this agent | **`traycer-host-start`** | `(null)` | agent |
| #2 — legacy CLI agent | **`sh`** | `(null)` | legacy agent |

```
 #33:
                 Name: traycer-host-start
       Developer Name: (null)
                 Type: agent (0x8)
           Identifier: 8.ai.traycer.host.agent
      Executable Path: Contents/Library/LaunchAgents/Traycer Host.app/Contents/MacOS/traycer-host-start
    Parent Identifier: 2.ai.traycer.desktop
```

**Signing changes nothing here** — the shipped notarized build recorded exactly what the ad-hoc probe predicted, from the `BundleProgram` basename. The defect is live in production today.

Note `Executable Path` already contains `Traycer Host.app`: the bundle directory was named correctly all along and BTM still ignored it, which is the same point the probe's v2 arm made.

## Scope, stated precisely

This changes the **child agent** row only. The parent app row (#32) already reads `Traycer` / `Traycer AI, Inc.` and is untouched — so the top-level Login Items entry was already correct, and what improves here is the expanded per-item name and the background-item notification text.

`Developer Name: (null)` on the child is **not** addressed by this PR, and now appears to be structural rather than a signing artifact: it is `(null)` on a correctly Developer-ID-signed, notarized release, while the parent carries the full attribution. That closes out the open "M1" question from the backlog — dev-signing does not populate it, so there is nothing to fix there.

PR #803, opened 2026-07-29
feat(gui-app): add persistent multi-project picker

### LINKED ISSUE #786: Feature: Add a persistent multi-project picker for chats
### Problem / motivation

Traycer currently starts a new chat with the last/default project folder. When users work across multiple codebases, switching projects requires adding another folder, removing the current one, and later re-adding folders they previously used. This creates repeated setup work and makes project switching feel destructive.

Users should be able to keep multiple project folders/codebases saved, choose which one or more are available to a specific chat, and separately control the default project for new chats.

### Proposed solution

Add a persistent project picker to the new-chat/composer experience.

### Suggested behavior
- Keep multiple project folders/codebases in a saved list instead of replacing the current folder.
- Let users select one or more saved folders to use in the current chat.
- Let users pin one folder as the default for new chats, independently of the current chat selection.
- Keep actions to add a folder and remove a saved folder.
- Show useful project context such as folder name, location/type, and current branch where available.
- Show the active selection and number of folders being used by the current chat.

### Acceptance criteria
- Adding a folder makes it available for future chats until the user removes it.
- A new chat starts with the pinned default folder.
- Changing folders for one chat does not change the pinned default.
- Users can select multiple folders for a chat and Traycer includes their paths as context.
- Switching between projects does not require removing and re-adding folders.

Reference mockup: https://claude.ai/code/artifact/b1105a9e-3f45-456a-8d78-57e97359b09b

### Alternatives considered

Continue using the current add/remove workflow or open a separate desktop window for each project. Issue #104 fixed folder isolation between desktop windows, but it does not provide a persistent project list or allow selecting multiple projects for one chat.

### Component

Desktop app

ReAVRafalSztaba:
## Summary

- keep saved projects separate from each chat's selected workspace folders
- let users switch the main project, opt into additional folders, and pin the default for new tasks
- seed blank and new drafts from the pinned project while preserving picker choices through draft minting
- keep composer context, mention roots, and launch data scoped to the chat's selected folders

This removes the destructive add/remove cycle when moving between codebases: saved projects remain available for future chats while each chat owns its current selection.

## UI preview

Persistent multi-project picker showing saved projects, the pinned default, and multi-folder selections

## Related issue

Closes #786

## Validation

- Manually verified the updated picker UI and project-selection workflow in the desktop app
- `pre-commit run --all-files` passed all available hooks, including the full Nx workspace check; the ShellCheck hook could not run because `shellcheck` is not installed locally
- Added and updated tests for saved-project persistence, per-chat selection, draft minting, picker controls, and mention roots

## Checklist

- [x] Pre-commit hooks pass (`pre-commit run --all-files`)
- [x] Commits are signed off (`git commit -s`) per DCO

### DIFF
--- a/clients/gui-app/src/components/epic-canvas/sidebar/new-conversation-modal.tsx
+++ b/clients/gui-app/src/components/epic-canvas/sidebar/new-conversation-modal.tsx
@@ -86,6 +86,7 @@ import {
   workspaceComposerCanStart,
 } from "@/lib/composer/workspace-composer-availability";
 import { effectiveWorktreeIntent } from "@/lib/worktree/effective-worktree-intent";
+import { resolvePrimaryPath } from "@/lib/worktree/resolve-primary-path";
 import { deriveWorkspaceMode } from "@/lib/worktree/workspace-mode";
 import { cn } from "@/lib/utils";
 import { ActiveHostWorkspaceControls } from "@/components/home/host-workspace-selector/host-workspace-selector";
@@ -1043,13 +1044,29 @@ function useLatestConversationSettingsSeed(): {
 }
 
 function useGlobalWorkspaceSnapshot(): LandingDraftWorkspaceSnapshot {
-  return useWorkspaceFoldersStore(
-    useShallow((state) => ({
-      folders: state.folders,
-      folderInfoByPath: state.folderInfoByPath,
-      primaryPath: state.primaryPath,
-    })),
+  // A brand-new conversation (no latest-conversation seed) starts with ONLY
+  // the pinned default project selected - the rest of the saved list stays
+  // available as unselected picker rows, mirroring the landing draft seed.
+  const folders = useWorkspaceFoldersStore((state) => state.folders);
+  const folderInfoByPath = useWorkspaceFoldersStore(
+    (state) => state.folderInfoByPath,
   );
+  const pinnedPath = useWorkspaceFoldersStore((state) => state.pinnedPath);
+  return useMemo(() => {
+    const pinned = resolvePrimaryPath(folders, pinnedPath);
+    const pinnedInfo =
+      pinned !== null && Object.hasOwn(folderInfoByPath, pinned)
+        ? folderInfoByPath[pinned]
+        : null;
+    if (pinnedInfo === null) {
+      return { folders: [], folderInfoByPath: {}, primaryPath: null };
+    }
+    return {
+      folders: [pinnedInfo.path],
+      folderInfoByPath: { [pinnedInfo.path]: pinnedInfo },
+      primaryPath: pinnedInfo.path,
+    };
+  }, [folders, folderInfoByPath, pinnedPath]);
 }
 
 function toTuiPlacement(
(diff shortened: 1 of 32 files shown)


anur4ag:
Hey, the design looks good overall. One concern: for users who have added many folders but only work with a few at a time, opening this picker could feel overwhelming.

Could we keep the currently selected folders visible and expanded, while placing the unselected folders inside a collapsed section? That would make the active context much easier to scan.

ReAVRafalSztaba:
> Hey, the design looks good overall. One concern: for users who have added many folders but only work with a few at a time, opening this picker could feel overwhelming.
> 
> Could we keep the currently selected folders visible and expanded, while placing the unselected folders inside a collapsed section? That would make the active context much easier to scan.

I understand but the picker displays the same amount of folders it did before the change, so If someone works on some projects currently they stay there, if not you can easily remove them from the list and add them back when someone needs to work on them again at least thats how I see it. Hiding them makes them not easily availble to be selected for the new chat. Its like a project picker, you work on something currently you have it in the list, and if you want the new chat/epic/task to be work on a different path then you just quickly select it.  

Let me know what you think.

anur4ag:
hi @ReAVRafalSztaba, apologies this took this long. I have made this PR (#1310) which solves the issue you were facing. Feel free to give feedback

ReAVRafalSztaba:
> hi @ReAVRafalSztaba, apologies this took this long. I have made this PR (#1310) which solves the issue you were facing. Feel free to give feedback

Hi @anur4ag 

I'm sorry, but how is it better? It makes much more sense to just pick a project that you want to work on currently than to add and remove and add and remove projects. Why are you so hung up on this concept? It doesn't work like that in any other AI coding app. In my opinion—and I'm 100% sure that if you asked users, they would say the same—my approach is much more straightforward and makes much more sense overall. 

I guess it does solve my issue but still it does not make sense for just switching between projects, i takes much more clicks to achieve the exact same thing with your approach. 

Just to make it make sense even more for you what my perspective is - When you work on multiple projects then you have a set of projects that you work on. In my approach you just add all the projects you work on and just select which one you want to work on at the moment, in your approach I add the project I want to work on and then I have to remove it and add another one, instead of just selecting the one I want. 

Seriously please reconsider just applying my approach.

anur4ag:
@ReAVRafalSztaba Thanks for the feedback. #1310  intentionally uses a different model: attached folders represent the current context, while inactive folders remain available under recent through progressive disclosure.

we chose this approach because it preserves the existing control-plane layout for users familiar with it and avoids overwhelming the picker with every saved folder. I appreciate the work you put into #803  and the suggestions you shared. we'll continue improving this flow without adding too many controls.

ReAVRafalSztaba:
Thanks for the reply @anur4ag 

It is your product you do as you want, but I still am 100% sure that if you asked your users which approach they liked better, the majority would choose a proper project picker. I don't really think you thought this through when it comes to the point you made now, which is overwhelming the picker with every saved folder. If you want to switch a project, you still need to open the recents folder, which you added, and it still opens up the list. You only open the picker when you actually want to switch the project. So if you want to switch the project, you always open the picker and you always open the recents. With your approach, it will actually be bigger and more overwhelming, not smaller.

But anyway its better then nothing. 

Please tell me when will the update land ? Because its really hard to work without being able to properly switch between projects.

anur4ag:
we did an RC for new release will do another with some more fixes (inc. this), then after testing release it with v1.2.0

ETA: 12-24 hrs

ReAVRafalSztaba:
@anur4ag okay thank you.

PR #835, opened 2026-07-30
fix(protocol): gate archive legend to host-enriched A2A listings

hdkshingala:
## Summary

The A2A list formatter always appended the `[archived]` legend line, even though versioned/CLI-parsed `agent.list` responses strip the host-only `archived` field before formatting reaches the CLI. This meant every CLI listing advertised archive-state semantics it could never actually represent, and archived rows were indistinguishable from unarchived ones in that output.

- `formatAgentListResponse` now computes `showArchived` from row-level enrichment presence (`"archived" in agent`) and threads it into both legend branches.
- Row markers continue to key off `isArchivedAgent` (`archived === true`); presence-gating the legend needs no separate change there, since an absent key can never be `true`.
- No schema changes — `listAgentsResponseSchema` and the versioned RPC family are untouched.

This is a compatibility fix for the archived-agent feature landing on the internal host side (`traycer_archive_agent` A2A tool); this PR is scoped to the shared protocol formatter only.

## Test plan
- [x] `bun run --filter @traycer/protocol test` — 106 files / 1468 tests passed
- [x] `bun run --filter @traycer/protocol compile` — passed
- [x] `bun run --filter @traycer/protocol lint` — passed
- [x] Prettier check on touched files — passed
- [x] New regression test: enriched listing with all-`false` archived rows still shows the legend (presence, not truthiness); stripped/versioned response shows neither marker nor legend


hdkshingala:
@coderabbitai approve

PR #904, opened 2026-08-02
fix(gui-app): highlight active terminal navigator item

tanveergill:
## Summary

Raw terminal sidebar rows used the record-backed artifact selector, which intentionally excludes terminal tiles. Use the existing active-tile selector so the terminal currently visible in the active canvas pane receives its navigator highlight.

Adds an integration regression test for the active terminal row.

## Related issue

None.

## Checklist

- [x] Affected build, compile, lint, and format checks pass via pre-commit
- [x] Targeted terminal-sidebar tests pass
- [x] Tests added where it makes sense


tanveergill:
@coderabbitai approve

PR #918, opened 2026-08-03
feat(protocol,gui-app): durable background cards for backgrounded commands

AmiteshwarRandhawa:
## What / Why

Codex decides on its own to yield a long-running exec to a background terminal, and keeps it alive past the turn that started it — through the next turn, and through everything short of the app-server dying. The real terminal can land minutes later. Until now a `command` block had no way to say any of that: it was folded into the activity group like any other command, force-finalized when its turn ended, and its terminal arrived to a card that had already claimed an ending.

This adds the wire markers and the GUI rendering for a backgrounded command's own durable card, and fixes the renderer routing that dropped its late terminal.

## Protocol

`commandBlockSchema` gains the same two markers `toolCallBlockSchema` already carries, with the same semantics:

- **`backgroundTask`** (nullable, defaults false) — sticky: survives every terminal path and reload, so the card stays standalone for the whole lifecycle instead of collapsing back into the activity group.
- **`stopped`** — the terminal outcome was an explicit stop, not a failure. The provider reports its own kill as a synthetic exit code, and rendering that as an error blames the command for something the host did.

`command.started` gains an optional `backgroundTask`, and the accumulator now **upserts** on it rather than blindly appending — a harness that learns about backgrounding only after the card is open (Codex decides at the parent turn's end) re-emits the event to stamp the marker, and appending would duplicate the card. The open block's `timestamp` (its elapsed anchor) and `status` are left alone.

`command.completed` gains `terminationReason` and `backgroundTask`, mirroring `tool_call.errored` / `tool_call.completed`.

`streamingDetachedBlockIds` now treats a background-marked `command` like a background-marked `tool_call`, so a clean `turn.completed` leaves it streaming instead of force-finalizing it — the block waits for its own terminal, which is the only frame that knows how it ended.

## GUI

- `shouldPromoteCommandSegment` mirrors the tool-call promotion gate (durable marker, plus the transient host `backgroundItems` set for a card whose marker hasn't arrived yet).
- `CommandSegment` grows a real card variant — elapsed heartbeat as the collapsed preview, a "Stopped" badge in place of the exit code when the host stopped it.
- **Detached-owner routing fix:** the renderer's detached router only recognized `subagent.*` and terminal `tool_call.*` events (Claude's background shapes). A parentless `command.completed` returned null, fell through toward the active turn, and was dropped — the card's timer ticked forever and only corrected on the next send, when the store re-derived from the host snapshot (live-repro on a 20s sleep that ticked past 2 minutes). `command.completed` is now routed exactly like a terminal `tool_call.*`: to its `parentBlockId` when it is a subagent child, otherwise to its own block's settled row, and mandatory when the terminal carries the background marker so it can never mint a duplicate card on an unrelated live turn.

## Deliberately not included

A per-item `stoppable` flag on the wire. `chat.subscribe` is frozen through 1.5, so adding a field to `backgroundItemBaseFields` is a breaking host→client change at every released version; it needs a new minor with frozen copies, not an additive default. Consequence today: on a Codex build below the stop-API floor the host lists the command but declines the stop, and the GUI cannot yet hide the button for it.

## Pairs with

This is the client half of the Codex background-work feature; the host half lives in an internal-repo PR (linked in a follow-up comment). Both are needed for the cards to appear.


AmiteshwarRandhawa:
Pairs with the internal-repo host PR: traycerai/traycer-internal#4788 (Codex background work — helper cards, background panel, and backgrounded commands). That PR's gitlink currently pins this branch head and will be re-bumped to the merge commit once this lands, so this one should merge first.

PR #923, opened 2026-08-03
fix(gui-app): retain resource monitor search

tanveergill:
## Summary
- retain the Resource Monitor search query when its popover is dismissed
- preserve the query after opening a matching resource so users can inspect successive matches
- add regressions for outside-click dismissal and resource selection

## Verification
- `bun run test src/components/resources/__tests__/resource-monitor-popover.test.tsx`
- `bun run compile`
- `bun x -y react-doctor@latest . --verbose --diff main --offline --no-score`
- pre-commit affected Nx build, compile, lint, and format checks


tanveergill:
@codex review

tanveergill:
@coderabbitai approve

PR #953, opened 2026-08-04
feat(gui-app): add translate-to-spanish action to selection popover

HarleyVasquezcom:
## Summary

Adds a "Traducir al español" action (Languages icon) to the transcript selection popover, next to the existing Quote action.

- `clients/gui-app/src/components/chat/quote/translate-to-spanish.ts` (new): builds the composer request (`Traduce al español el siguiente texto:` + quoted selection) and appends it via the composer's `replaceDraft` focus-end path.
- `quote-selection-popover.tsx`: renders a second action button reusing the existing Quote button styling; consumes the same mouseup-selection snapshot contract.
- Tests: new unit coverage for the request builder/append plus a popover test for the new action.

## Verification

- GUI compile (`tsc -b`) passed
- ESLint passed (`--max-warnings 0`)
- quote popover + translate-to-spanish suites: 77/77 tests pass

---

Commit signed off per the repo DCO requirement.

### DIFF
--- a/clients/gui-app/src/components/chat/quote/__tests__/quote-selection-popover.test.tsx
+++ b/clients/gui-app/src/components/chat/quote/__tests__/quote-selection-popover.test.tsx
@@ -140,6 +140,24 @@ describe("QuoteSelectionPopover - quote action", () => {
   });
 });
 
+describe("QuoteSelectionPopover - translate action", () => {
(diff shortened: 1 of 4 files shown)


pranshugupta54:
Thanks, but passing on this. A hard-coded single-language prompt button is too narrow for the shared selection popover, and the same result is typing one sentence after Quote. Would consider a generic configurable prompt-action instead. Closing.

PR #959, opened 2026-08-04
fix(gui-app): keep an activity group's reasoning header stable, and its summary honest

hdkshingala:
Follow-up to #899. The accumulated activity group shipped with a summary that could claim a duration it did not have, a header that appeared and vanished as the group's shape changed underneath the reader, and rows whose only affordance was a click target.

## Summary

- **Honest thinking phrase.** Three states rather than one: a still-streaming block contributes its presence only, so a group mid stream reads *thinking*, a finished block with no duration reads *thought*, and only a known duration reads *thought for 1m 4s*. Previously a partially-streamed group could report a total it had not measured.
- **The headerless rule is latched on a header the reader actually saw.** `hidesSoleReasoningHeader` is derived from group shape, and shape flips **both** ways — four removal paths across two files. A reverse flip stripped a visible header and, because the body-shown rule includes `headerless`, system-expanded a settled trace: the #597 discontinuity inverted. The latch lives in the activity-group open store so it survives a tab remount and LegendList row recycling, and is written from an effect gated on children actually being on screen *and* a reasoning header actually rendering.
- **The suppressed header stays keyboard-reachable.** `sr-only focus:not-sr-only`, with an action name that follows its state instead of contradicting it, and a focus handoff to the reasoning body when completion removes the control out from under the reader.
- **Promoted and revealed rows land somewhere named.** Focus moves to the row's own trigger, the row scrolls into view, and the revealed child opens rather than landing shut.
- **One shared `RowTrailingCaret`** between expandable and promoting rows, so a row that promotes reads the same as a row that expands.
- **Live elapsed is inline**, beside the pulse, matching the background-command card — instead of splitting the timer onto a second footer row. The old behaviour came from a variant gate, so the same command showed its timer inline on a card and stacked in a group.
- **The streaming footer is just a progress line**, and renders nothing when there is no progress.

### Find projection

The suppressed header and the elapsed labels are `data-find-skip` chrome. The renderer drops the find *anchor* whenever the projection drops the *unit*, so a counted match with nowhere to paint cannot arise.

**Known, accepted asymmetry:** the find projection is rebuilt from the model and cannot see the latch, so in the latched state the renderer shows a header the projection does not index. That is a suppressed match, not a phantom one.

## Design notes

The latch is the third design tried; the first two were wrong in ways worth recording:

| Attempt | Why it failed |
| --- | --- |
| A `shrunk` flag on the model | Blind to removals upstream of the builder, and could not tell "shrank under the reader" from "arrived this shape" — it restored the duplicate header it was meant to remove |
| A component `useState` latch | Died with the mount while the open store survives tab remounts, and latched headers nobody had seen |
| **Store latch, uncapped** | Observes the *consequence* rather than enumerating causes, and seeds correctly on a cold mount |

`headedIds` is deliberately **uncapped**, unlike `openIds`. A FIFO cap there is not a cap, it is amnesia: evicting the oldest entry erases the fact that a header was once on screen, and the header silently vanishes.

## Verification

- lint, format and compile clean
- full gui-app suite green: **1034 files / 10580 tests**, 1 skipped
- mutation-probed the headerless latch, the find projection and the focus handoff

**Note for future probes:** the two focus guards are redundant by design — one says "the control went away", the other "the reader did not move focus themselves". Removing either alone is unobservable, so a single-guard mutation *survives*. The tests discriminate on the pair, and say so inline. Please don't read a surviving single-guard mutant here as a coverage gap.

## Live acceptance limitation

Not yet exercised in a running Electron window. Timer placement, hover carets, focus ring and reveal scroll are covered by automated assertions only; this PR does not claim a pixel-level acceptance run.


hdkshingala:
All seven findings are addressed in `2ab185bd`. Two changed behaviour; five were test hygiene. Notes on the three where I did something other than what was suggested.

### Partial thinking duration (Codex P2) — real hole, different fix

The observation is right and worth fixing: a group holding one measured block and one finished-but-unmeasurable block reported the partial sum as a total.

Two corrections though. **This is not a regression** — the parent gated on `thinkingDurationMs <= 0` and produced exactly the same string, so the duration-less `Thought` the comment attributes to it never happened. And **the suggested remedy would reintroduce a defect this PR exists to fix**: requiring every duration to be known before presenting the sum walks the label backwards, `thought for 5s` → `thought`, the moment an unmeasurable block joins. That reverse step is what `thinkingCompleted` was added to prevent.

It is also reachable live, not just in persisted history — `completedDurationMs` returns null for any block whose timestamp is the turn-end, so **interrupted, superseded and errored** blocks all land here. Interrupt a turn mid-thought and resume, and the group holds one of each.

So the sum is presented as what it is, a floor:

| Group | Reads |
|---|---|
| one block, 5s | `Thought for 5s` |
| 5s + unmeasurable | `Thought for 5s+` |
| 5s + unmeasurable + 6s | `Thought for 11s+` |
| unmeasurable only | `Thought` |

A floor only grows, so monotonicity holds. The `+` needs a measured block *and* an unmeasurable one — two segments — and a two-segment group is never headerless, so it cannot break the `hidesSoleReasoningHeader` equality. That is asserted, not argued.

### `onHeaderBlur` during removal — agreed, and more severe than "minor"

This one would have been **dead in the shipping browser**. jsdom dispatches no blur when a focused element is removed; Chromium does. So removal itself cleared the latch before the effect ran, and the handoff never fired in Electron — with every test green, because jsdom never produces the event that breaks it.

Fixed by clearing only on a blur that names where focus went: a reader who tabbed away hands focus to an element, removal hands it to nothing. The new test fires the blur jsdom won't, so it fails for the code rather than the environment.

### Two suggestions not taken as written

**`closest("button")` does not transfer** from the command-segment suite. A streaming tool row takes the non-expandable `SegmentRow` branch, which is a `div`, so `closest("button")` returns `null` and the assertion fails for the wrong reason. Added a `data-row-header` marker to all three branches instead, plus `data-row-caret` for the caret assertions — same intent, one anchor that means the same thing whichever branch rendered.

**The `item === undefined` guard is rejected by lint.** This repo does not set `noUncheckedIndexedAccess`, so the element type excludes `undefined` and `@typescript-eslint/no-unnecessary-condition` refuses a comparison the type system believes impossible — while the runtime disagrees. The range check moved to `timeline.length`, which both agree on.

### One item was a vacuity, not a flake

Pinning the clock on *"omits the heartbeat once the call completes"* mattered more than the intermittency suggested: `queryByText("0s")` also returns null when a counter **is** rendered and the clock has ticked to `"1s"`, so a regression that kept the heartbeat alive would have passed. It now matches any counter.

---

Both behaviour changes are mutation-probed — dropping the floor marker kills 2 tests, clearing the latch on any blur kills 1. Full gui-app suite green at 1034 files / 10587 tests.

hdkshingala:
Fixed in `a23a9553`. This one was real, and it is the same defect class the PR exists to fix, reached through the latch's own blind spot.

`LiveActivityWindow` keeps its rows in the DOM for the whole `LIVE_ACTIVITY_WINDOW_EXIT_MS` — `shown` going false starts the fold, it does not take the children off screen. The latch gated on `shown`, so a group that settles in the **same update** that adds a second segment turned the window off while the newly headed reasoning row stayed visible for another 300ms, and recorded nothing. A later shrink then stripped a header the reader had seen and unfolded a settled trace.

The latch now gates on whether the window's children are mounted.

The exit timing moved into `useLiveActivityWindowMounted` so `shown`-vs-mounted has one owner. Worth noting two judgement calls in how:

- **Called twice rather than lifted into a prop.** My first attempt made `mounted` a required prop of `LiveActivityWindow`, which is cleaner in the abstract but forced `mounted` into all 13 renders in that component's own test suite and moved its exit behaviour out of the tests that own it. Both call sites take the identical `shown` value in the identical commit, so they enter and leave together — and the latch is monotonic, so a tick either way only ever adds.
- **Its own module**, because `react-refresh/only-export-components` refuses a hook exported beside a component.

**Probe:** gating back on `liveWindowShown` fails the new test, before and after that refactor. The test asserts the child header is genuinely on screen mid-exit *before* advancing timers, so it cannot pass against a window that had already torn its rows down.

Full gui-app suite green at 1034 files / 10588 tests.

hdkshingala:
Both P2s were real. Fixed in `a2e10dd6`.

### The latch was orphaned when a group's first member left

This is the one I got wrong in my own reasoning, not just in the code. The comment on that latch argued about `[reasoning, X] -> [reasoning]`, where reasoning leads and the group id happens to hold still — and concluded the id was stable. It isn't: the id is `deriveActivityGroupRenderId(segments[0].id)`, so `[command, reasoning]` losing its command **renames the group**, and a group-keyed latch is lost across exactly the move it exists to survive.

`headedIds` now holds reasoning **segment** ids. Every reasoning segment in a headed group is marked rather than just one, because which of them ends up alone is decided later by which *other* member leaves — the survivor is not knowable while its header is on screen.

**Probe:** keying back on `group.id` now fails **five** tests, three of which predate this round. The new test asserts the rename actually happens before relying on it, so it cannot pass against a fixture where the ids coincide.

### The ambiguous blur

Correct, and it is the exact counter-case to the previous commit — I traded one bug for the other. `relatedTarget === null` is both Chromium's removal blur *and* a click on a non-focusable surface, and they are indistinguishable at dispatch time.

A layout effect now re-reads real focus ownership after every commit while the control exists, and preserves the last reading on the commit that removes it. The blur handler keeps only the unambiguous case, so:

| Event | Latch | Outcome |
|---|---|---|
| Tab/click to another element | cleared by `relatedTarget` | no handoff |
| Click a non-focusable surface | cleared by the next commit's sampler | no handoff |
| Control removed | sampler bails, last reading preserved | handoff fires |

Implementation note: it samples via `querySelector` on the wrapper rather than a ref threaded into `ReasoningHeader`. The React compiler lint rule rejects the ref-as-prop form outright — *"Cannot access refs during render"* — and it also poisoned the sibling `props.onToggle` / `props.onFocus` reads. The DOM query is simpler anyway, and "control gone → preserve the reading" falls out of `querySelector` returning null.

### Role query

Applied.

---

Full gui-app suite green at 1034 files / 10590 tests.

PR #969, opened 2026-08-04
fix(gui-app): isolate shifted shortcut hints

anur4ag:
## Summary

- prevent an unowned Cmd+Shift hold, including macOS screenshot shortcuts, from revealing ordinary canvas and header leader hints
- preserve Cmd-based inner-tab navigation and the Cmd+Shift model-profile picker when two or more profiles own that shifted chord
- spend pending and visible shifted sessions when their exact owner disappears, so scope changes cannot leak stale hints

## Verification

- keybinding visual suite: 34/34 passed
- focused provider/context/chord suites: 76/76 passed
- GUI compile passed
- React Doctor changed-files scan passed
- pre-commit safety, build, compile, lint, format, and DCO checks passed

The internal traycer-internal gitlink pin will follow after this OSS change lands, per repository policy.


anur4ag:
@coderabbitai approve

PR #983, opened 2026-08-04
docs(gui-app,shared): correct the catalog carve-out comments

hdkshingala:
Follow-up to #977. Comment-only.

Both call-site comments still describe the behaviour that PR's review round
removed — "marked stale and picked up at the intent edges". The shipped code
skips the harness-catalog keys entirely: it does not refetch them **and does
not mark them stale**.

That distinction is the whole point of the fix, so a comment asserting the
opposite is actively misleading to the next reader. `Query.isStaleByTime`
returns `true` on `state.isInvalidated` *before* it consults the time budget
(query-core 5.101.4, `query.js:134`), and `shouldFetchOnMount` routes through
it — so an invalidated entry refetches at the next mount regardless of
`staleTime: Infinity`. Marking these keys would have moved the ~14-harness
provider-CLI burst from the recovery sweep to the user's next picker open,
rather than preventing it.

No executable change.


hdkshingala:
@coderabbitai review

PR #994, opened 2026-08-05
fix(gui-app): restore sidebar item separators

anur4ag:
## Summary

Restore the missing separator lines between tiled sidebar panels and vertically split canvas panes.

- Add `before:absolute` to the vertical in-flow drag handle pseudo-element.
- Put the expanded-run bottom border on the run wrapper so it separates adjacent runs.
- Assert `before:absolute` in the resize handle regression test.

## Root cause

PR #924 reintroduced the 1px divider as a `::before` pseudo-element. The horizontal branch had `before:absolute`, but the vertical branch did not, so its static inline pseudo-element ignored the height and inset styles. In a multi-panel expanded run, each section is the only child of its split wrapper, so `last:border-b-0` always wins and the section border never paints.

## Verification

- Focused GUI suites: 53 tests passed across the sidebar and resize-handle files.
- Pre-commit affected build, compile, lint, and format checks passed.
- `bun.lock` is unchanged.
- Commit includes the required DCO sign-off.

### DIFF
--- a/clients/gui-app/src/components/epic-canvas/canvas/__tests__/resize-handle.test.tsx
+++ b/clients/gui-app/src/components/epic-canvas/canvas/__tests__/resize-handle.test.tsx
@@ -124,6 +124,9 @@ describe("<SplitResizeHandle />", () => {
     expect(className).toContain("h-1");
     expect(className).toContain("before:h-px");
     expect(className).toContain("before:top-1/2");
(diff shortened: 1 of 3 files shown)


anur4ag:
@coderabbitai review

anur4ag:
@coderabbitai review

anur4ag:
@coderabbitai approve

PR #1008, opened 2026-08-05
fix(gui-app): prevent tooltips from swallowing folder-picker clicks

### LINKED ISSUE #466: Folder picker in "Start a new chat" becomes unclickable — stuck hover tooltip intercepts clicks
### Component

Desktop app

### Version

1.1.6

### OS / platform

macOS (arm64)

### What happened?

In the workspace/folder picker inside "Start a new chat" (both from the start page and the bottom workspace selector), I cannot select a different project folder. Clicking a folder row (e.g. `mantis`) does nothing — the mouse cursor even changes to a text-caret over the rows instead of a pointer, and no selection happens.

The likely trigger: a hover tooltip — **"Worktrees require a Git repository. This folder can only run Local."** — gets stuck rendered on top of the folder list and never dismisses. Once it's stuck, the whole picker becomes unresponsive to clicks. Opening it in a new window doesn't help either.

Only workaround found: reload the window (Cmd+R) or fully quit and relaunch the app.

This looks like the same underlying problem as #446 ("Cannot change project in 'start page'"), with an additional detail on the trigger (the stuck tooltip overlay intercepting clicks).

### Steps to reproduce

1. Run Traycer desktop
2. Open "Start a new chat" and click the workspace/folder selector
3. Hover a non-Git folder so the "Worktrees require a Git repository. This folder can only run Local." tooltip appears
4. The tooltip stays stuck over the list; clicking any folder row no longer registers

### Logs / output

```shell

```

anur4ag:
## Summary

- Disable hit-testing on Radix Popper wrappers that contain label-only tooltips, so covered folder-picker rows receive clicks.
- Keep the tooltip content-level `pointer-events-none` class and document why both layers are required.
- Add jsdom regression coverage for the authored selector and the wrapper DOM Radix renders.

## Related issue

Closes #466

## Validation

- `bun run test -- src/components/ui/__tests__/tooltip-hit-testing.test.tsx` (2 passed)
- `bun run test -- src/components/ui/__tests__/hover-preview-surface.test.tsx src/components/home/host-workspace-selector/__tests__/folder-controls.test.tsx src/components/home/host-workspace-selector/__tests__/workspace-folders-refresh.test.tsx` (69 passed)
- Targeted Prettier and ESLint checks passed.
- Commit pre-commit hook passed affected build, compile, lint, and format checks.

## Checklist

- [ ] `bun run build`, `bun run lint`, and `bun run test` pass
- [x] Code is formatted (`bun run format` equivalent check passed)
- [x] Pre-commit hook passed on commit
- [x] Commit is signed off (`git commit -s`) per the DCO

### DIFF
--- a/clients/gui-app/src/components/ui/__tests__/tooltip-hit-testing.test.tsx
+++ b/clients/gui-app/src/components/ui/__tests__/tooltip-hit-testing.test.tsx
@@ -1,0 +1,86 @@
+/**
+ * Regression coverage for traycerai/traycer#466 (and the click half of #446):
+ * a label-only tooltip swallowing clicks meant for whatever it covers.
+ *
+ * `TooltipContent`'s `pointer-events-none` is not enough on its own - Radix
+ * Popper renders that content inside a same-size positioning wrapper that keeps
+ * `pointer-events: auto`, and THAT wrapper won hit-testing. In the workspace
+ * folder picker (28px rows, `side="top"` tooltips) it landed on the row above,
+ * so clicking that row did nothing. The fix is one rule in `index.css`.
+ *
+ * jsdom has no layout and cannot hit-test, so the two halves are pinned
+ * separately and joined by construction: the CSS rule is read out of
+ * `index.css`, and its selector - the exact string, not a copy - is then
+ * matched against the wrapper a real Radix tooltip renders. Deleting the rule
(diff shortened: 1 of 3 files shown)


anur4ag:
@coderabbitai approve

PR #1009, opened 2026-08-05
feat: chat-anchored monitors & shells with same-turn delivery wire

AmiteshwarRandhawa:
## What

The chat becomes the home for its monitors and shells, and the queue-item wire gains same-turn delivery fields.

**Protocol** (`chat.subscribe`, still-unshipped 1.6 line — no version ritual needed, verified absent from all release tags):
- The managed-command queue item gains `delivery`/`targetTurnId` and a `steering` status, all defaulted so rows written by earlier builds rehydrate as the next-turn fallback.

**GUI:**
- **Floating top-right menu on the chat tile** — present exactly while the chat owns a command in any state (never yanked while open); badge: quiet / running count / attention count (attention only on unexpected exits, beats running, acknowledged by opening the menu, re-arms on re-failure).
- **Menu rows**: all the chat's commands, running first; stop/start/delete with the destructive-delete confirmation; click opens the output window; **drag a row out to open — or move — its output window at the drop target** (rides the canvas dnd machinery; move-vs-open resolves by content id).
- **Epic sidebar "Monitors & Shells" list removed** — persisted layouts rehydrate safely (unknown panel ids are filtered, not layout-resetting).
- **Background panel unified**: managed commands are ordinary rows in the one list (kind icons carry the monitor/shell distinction; the harness's own "monitor" kind keeps its label and screen icon), the "Not stopped by Stop all" disclaimer is gone because **Stop all now genuinely stops everything listed** — one aggregated action, single pending state, single partial-failure toast. Managed-row stop no longer gated on the chat stream.
- **Output window follows the Terminal font settings** (family + size, via a shared resolution hook also adopted by the terminal tile and appearance panel).
- Stale-list / host-too-old notices re-homed into the menu.

## Review

Built by two independent implementation rounds, then an adversarial review (probe tests, not inspection) that confirmed 8 defects — including drag-out being a complete no-op due to popover unmount killing the dnd payload, and the sidebar-layout upgrade reset — followed by a fix round in which every fix landed with the test that would have caught it, proof-of-red verified.

## Validation

Targeted suites green (drag/pointer-gesture suite included), gui-app compile, ESLint, Prettier, react-doctor all clean. Known residual: drag-out deserves one manual browser pass (jsdom can't fully prove the pointer gesture).


AmiteshwarRandhawa:
Review feedback addressed in 60272aaf:

- **Stop all per-half gating** (Codex P2 + CodeRabbit): the button is offered when either half can act and each half is sent on its own capability — a reconnecting chat keeps the one-click stop for managed commands, and a press during the managed fan-out can no longer re-send the set. Button-level assertions added alongside the existing row-level test, including the disabled-only-when-neither-half-can-act case.
- **Stop-all hook fail-fast** (CodeRabbit): when no host client can be built, the mutation throws the real `hostClientUnavailableError` once instead of manufacturing N identical rejections that collapse into an uninformative count.
- **Role query** (CodeRabbit): the output-tile timeline test now binds to the exposed `role="log"`.

Skipped: the `toastFromHostError` fallback-argument suggestion — the hook doesn't call that helper; the single-toast-with-message design stands.

AmiteshwarRandhawa:
Second round addressed in 4d466aa3:

- **Steering chip state** (Codex P2): a managed-command chip in the handover window now labels itself "Delivering" — the locked controls and closed cancel lever explain themselves. Test added asserting the label, aria-busy, and the absent cancel.
- **Role queries** (CodeRabbit): the new Stop all assertions bind to the button's role and accessible name, asserting `HTMLButtonElement.disabled`.

AmiteshwarRandhawa:
Pushed 03536019 (with the `main` merge):

- **Placement change per product-owner review**: the Monitors & Shells trigger moved from the floating top-right overlay into the workspace-controls row below the composer, left of the context-usage chip — in-flow, natural tab order, popover opens upward; the overlay wrapper and its minimap special-casing are deleted.
- Merge of `main` resolved.
- Fixed four suites that were red on the branch prior to the merge (renders missing the tab-host provider / mock exports the panel now requires). Full gui-app suite: 11,126 tests passing.

AmiteshwarRandhawa:
Third round addressed in b46b38d6:

- **Delete confirmation survives its popover** (Codex): same focus-outside guard as the notifications bell — the portaled dialog no longer dismisses the menu out from under itself.
- **Popover capped to the viewport** (Codex): max-height + scroll, so many finished shells can't push rows off screen.
- **Per-row drag ids** (Codex): a useId occurrence key, same pattern as the active-agent rows, so the same chat open in two tiles can't cross drag wires.
- **Endings that arrive while the menu is open count as seen** (Codex): acknowledgement now tracks what is rendered, not just the open transition; the re-arm test was split into seen-while-open and re-arm-after-close cases.
- **Stop all: one button, one in-flight state** (Codex): disabled while anything it started is still running — no resubmitting the finished half.
- **CodeRabbit**: pulse label follows the visible "Delivering" state; badge counts joined the trigger's accessible name (extracted helper, no nested ternary); native disabled assertion; unavailable-host coverage for the stop-all hook.

Skipped, with reasons: the `useEffectiveTerminalFont` rename — the `use<Namespace><Verb><Noun>` convention governs the RPC query/mutation hooks; this is a settings resolution and the proposed rename trades clarity for conformance.

AmiteshwarRandhawa:
On "keep the unsupported-host notice reachable" — deliberately not taken. A host too old for `managedCommand.subscribeList` predates the managed-command subsystem entirely (no agent tools either), so no monitors or shells can exist on it: the absent trigger is truthful, per the presence-equals-existence rule, and an always-rendered unavailable affordance would put a permanently dead button on every chat of every older host. The genuine reachable case — the menu held open across a swap to an older host — already renders the too-old notice inside the popover.

AmiteshwarRandhawa:
**Architecture change (product-owner directed): managed commands now ride the chat stream; the epic list stream is deleted** (3a85baf1 here, be467ff339 internal).

`managedCommand.subscribeList` existed to serve a global sidebar panel that no longer exists; every remaining consumer is chat-scoped. chat.subscribe snapshots now carry `managedCommands` (`.default([])` — an old host and an empty chat both truthfully read as none, so consumers never null-check and the menu's host-too-old branch is deleted as unrepresentable) plus a dedicated whole-set `managedCommandsChanged` frame. Both wire lines are unshipped, so no versioning ritual applies.

**This resolves the cross-host P1 by construction**: chat.subscribe is tab-host-bound, so a chat tile reads its own host's commands — the mixed shape (app-wide stream feeding a tile-bound control) no longer exists. The list mount/store/registry/factory and stale-list plumbing are deleted; a future global panel's re-entry path is documented where the method lived (the epic-activity dual-plane is the intended rail for cross-host ambient summaries).

Known narrowing: an output window whose owning chat has no live session falls back to its persisted tab name until one opens. Follow-ups noted for a later PR: an exhaustiveness guard on the chat-stream client's frame switch (its absence let a new frame kind be silently dropped — caught and probed during this migration), and the output tile's capability check reading the app-wide client (`useStreamMethodSupportFor` exists for the per-tab form).

AmiteshwarRandhawa:
On "read resource stats from the tile-bound host" — valid, deferred to the cross-host follow-up alongside its two siblings (the output tile's capability check, and the chat-stream client exhaustiveness guard). The command list itself is now tab-host-bound by construction; the resource chips still join against the app-wide resources registry, so on the rare cross-host tile a chip goes ABSENT (a foreign command id misses the active host's snapshot) — which is the resources design's stated preference over a wrong or zero readout, and the readouts are opt-in. Re-scoping the resources stream per host reshapes a surface the app-level resources popover also rides, so it belongs to the resources work package, not this PR.

AmiteshwarRandhawa:
On "key managed-command chat slices by host" — refuted after verification, no change. The premise ("cross-host clones retain the same epic/chat in tabs bound to different hosts") is not a reachable state: clone-not-migrate creates a SIBLING chat with a new id on the target host (`clone-chat-on-host-switch.ts` → `openNewChatInActiveTile`/`createChat`; its doc comment says exactly this), and chat tabs are host-bound for life, so `(epicId, chatId)` uniquely determines one host and the chat record itself carries that `hostId`. The registry's dispose-on-scope-mismatch handles sequential scope changes (user/transport rebind over time), not two concurrent hosts owning one chat id. The managed-commands slice inherits the same keying as the transcript, composer, and queue — if this premise were reachable, the entire chat surface would thrash, not just this selector.

Issue #1003, opened 2026-08-05
Dictation failed


intentodesign:
### Environment

| Field | Value |
|---|---|
| App Version | 1.1.9 |
| Platform | win32 (x64) |
| Host | 1.1.9 (ready, pid 25512) |
| Electron | 42.7.1 |
| Chrome   | 148.0.7778.280 |
| Node.js  | 24.18.0 |
| Support Report | `rpt_60d8bfb1b2304bb583ec9607c9b7d33f` |

### What happened?

Area: Dictation

### Steps to reproduce

1.
2.
3.

hdkshingala:
Thanks for the report — the diagnostic bundle came through and told us a lot.

Your logs show the same pattern as another recent report (#945): dictation failed because the app's connection to the local Traycer Host service was unstable at that moment. Dictation runs the speech recognizer inside that host process on your machine, so when the link drops, dictation fails with the error you saw. In the half hour before your report the connection dropped and reconnected repeatedly while the host was under heavy load. It doesn't look like a microphone problem — though if it ever fails with no other errors around, it's worth confirming Windows Settings → Privacy & security → Microphone → "Let desktop apps access your microphone" is enabled.

What should help right away: if other things are erroring at the same time (models failing to load, agent errors), give the app a moment to reconnect — or restart it — then retry dictation. Fewer concurrent agent sessions will also reduce the drops on a busy machine.

On our side, yours is the second report of this exact pattern, so we're treating it as a real recurring bug, not a one-off. We've already filed work to make dictation report its precise failure cause, and the next release includes changes that reduce the background load that was destabilizing the connection.

hdkshingala:
Following up: the reporting work is merged (#1017, with a matching host-side change).

Yours was the second report of this exact pattern, which is what moved it from a one-off to a real gap. Until now every dictation failure produced the same generic message, and several of the failure paths wrote no log entry at all — so both reports had to be diagnosed by elimination rather than read directly. Each failure now names its own cause: host connection already down, connection lost mid-session, an error returned by the host, the microphone failing to open, permission refused. That code is filled into the "Report issue" form automatically, so the next report of this shape says which path failed. The host also logs when a dictation session opens and closes, so an attempt that never reaches the speech recognizer is visible rather than silent.

To be clear on scope: this doesn't fix the failure itself. Dictation failed because the app's connection to the local host service was unstable, and that's still open — this makes the next occurrence legible. It's merged but not in a released build yet.

[...]

Issue #945, opened 2026-08-03
Dictation failed


flivecompany:
### Environment

| Field | Value |
|---|---|
| App Version | 1.1.9 |
| Platform | win32 (x64) |
| Host | 1.1.9 (ready, pid 2160) |
| Electron | 42.7.1 |
| Chrome   | 148.0.7778.280 |
| Node.js  | 24.18.0 |
| Support Report | `rpt_57a46e64e5ed4f43a20c656cf522a5ec` |

### What happened?

Area: Dictation

### Steps to reproduce

1.
2.
3.

hdkshingala:
Thanks for the report — the diagnostic bundle came through and was genuinely useful.

From the logs, the most likely cause is that dictation lost its connection to the local Traycer Host service at the moment you used it. Dictation runs the speech recognizer inside the host process on your machine, and in the half hour before your report the app's link to that host was repeatedly dropping and reconnecting, which would make dictation fail with exactly the error you saw. It doesn't look like a microphone-permission problem — but if it happens again with no connection trouble, it's worth confirming Windows Settings → Privacy → Microphone has "Allow desktop apps to access your microphone" enabled.

Two things that should help right away:
- If you see other host errors around the same time (things failing to load, agents erroring), give it a moment to reconnect — or restart the app — and dictation should work again.
- Closing a few heavy applications can help too; the host service was under significant load at the time.

On our side: current dictation error reporting doesn't pinpoint the exact failure step well enough, so we're filing a ticket to improve that — future reports like yours will then tell us precisely what went wrong. We're also looking at the host stability issues from your other reports (#955, #964) — they may share the same underlying condition.

hdkshingala:
Following up on the reporting improvement mentioned above — it's merged (#1017, with a matching host-side change).

Until now every dictation failure looked the same from outside the app: one generic message, and several of the failure paths wrote no log entry at all. That's why your report had to be worked out by elimination rather than simply read. Each failure now names its own cause — host connection already down, connection lost mid-session, an error returned by the host, the microphone failing to open, permission refused — and that code is filled into the "Report issue" form automatically, so a future report identifies the path without anyone having to infer it. The host also records when a dictation session opens and closes, so an attempt that never reaches the speech recognizer shows up as such instead of as silence.

What this doesn't do is fix the failure you hit. Yours failed because the app's link to the local host service was dropping repeatedly, and that remains open — this only makes the next occurrence readable. It's merged but not in a released build yet.

[...]

PR #1017, opened 2026-08-05
feat(gui-app): give a dictation failure a class it can be triaged by

hdkshingala:
## Summary

- Route every `use-voice-dictation` failure through one `fail()` call that
  stamps a `DictationFailureClass` and the readiness facts around it.
- Reset the per-attempt diagnostic refs at the top of `start()`, before any
  failure path can read them.
- Carry that class into the Report Issue prefill as a stable app-defined
  code, keeping the raw error text in the private diagnostics.

## Root cause

Two field reports of "Dictation failed" (#945, #1003) could only be
root-caused by elimination. All ten failure paths in the hook set the same
generic message, five of them logged nothing at all, and the issue the user
filed carried no code - so nothing in a support bundle said which path had
failed, only that one had.

## What the log line answers

The class alone does not separate "the attempt never reached the host" from
"the host answered with an error", so the line also carries the socket's
last connection status, whether the session reached `ready`, whether the
audio graph came up, and how long recording had been running. It is emitted
from inside `fail()` and *before* `teardownAll()`, which clears the very
refs it reports - logging after teardown would describe the torn-down
session rather than the failing one.

The two paths that already had their own `warn` fold into that line rather
than adding a second, so a failure is exactly one line, always classified.

`start()` resets `recordingStartedAtRef` and `lastConnectionStatusRef`
before the not-connected branch, so a retry cannot report the *previous*
attempt's connection status or a still-growing time-since-recording. `null`
now honestly means "this attempt never got that far" instead of "stale".

## Report Issue prefill

The class is app-defined and fixed, so it is safe in the public issue body;
the error text is not, and stays in `privateDiagnostics`. `captureDictationFailure`
deliberately does not call `captureException` - most dictation failures are
environmental (no microphone permission, host link down), not defects, and
they would drown the issue stream.

## Analytics

Unchanged, deliberately. The new fields travel on a separate `details`
argument rather than through `error`, which feeds blocker classification -
enriching the log must not silently reclassify the funnel.

## Verification

- 21 tests across `use-voice-dictation.test.tsx` and
  `report-issue-error-capture.test.ts`, including one that pins a retry to
  *this* attempt's status rather than the last one's.
- Each new assertion proven non-vacuous by reverting the change under it and
  confirming the test fails.
- `bun run compile` and `eslint --max-warnings 0` clean on the OSS graph.

## Scope

Relates to #945 and #1003; it does not close either. Those attempts failed for
reasons upstream of dictation - this makes the next one legible in a support
bundle instead of an absence.

No protocol change. The host-side counterpart (open/close lines for the
`speech.dictate` session) lands separately.


hdkshingala:
@coderabbitai approve

Issue #1001, opened 2026-08-05
Could not start Traycer Host


ddefaveri:
### Environment

| Field | Value |
|---|---|
| App Version | 1.1.9 |
| Platform | win32 (x64) |
| Electron | 42.7.1 |
| Chrome   | 148.0.7778.280 |
| Node.js  | 24.18.0 |
| Support Report | `rpt_b9911e7d4e4b47b49e5e7b7bce3defd8` |

### What happened?

Area: Host startup

Traycer Host could not start.

### Steps to reproduce

1.
2.
3.

hdkshingala:
Thanks for the report — we pulled the diagnostics you sent and can see what happened.

Two separate things stacked up:

1. The first "Could not start Traycer Host" message appeared prematurely — on a fresh install the app currently gives the host a flat 60 seconds, and it fired before sign-in had even happened. We've already fixed this (#868) and it ships in the next release.

2. The real blocker: the host **installed** correctly, but Windows refused to **start** it. Traycer launches its background host through a Windows Scheduled Task (`\Traycer\Host`), and every start attempt failed with the same error — the host process never launched at all. This is almost always one of:
   - antivirus blocking or quarantining the small launcher script under `%USERPROFILE%\.traycer` (it's a `.vbs` file, which some AVs flag),
   - Windows Script Host disabled by policy (common on hardened machines),
   - the `\Traycer\Host` task being disabled or blocked in Task Scheduler.

To get unblocked:
1. Open **Task Scheduler** → Task Scheduler Library → `Traycer` → `Host`, and try **Run** on it manually. If Windows shows an error there, that's our culprit — we'd love to know what it says.
2. Add an antivirus exclusion for `%USERPROFILE%\.traycer`, then retry from Traycer, or run `%USERPROFILE%\.traycer\cli\bin\traycer.exe host doctor` in a terminal and follow what it reports.

We're also improving this path on our side: the next release makes the actual Windows error visible instead of a generic failure, and `host doctor` will detect the disabled-Script-Host case explicitly. We'll follow up here when it's out.

hdkshingala:
Correction, and an apology — I gave you the wrong information.

I said the Doctor check for the disabled Windows Script Host would arrive in the next release. It was already there. That check shipped in **v1.1.9**, the build you were running when you filed this, so you could have run it at the time instead of waiting:

```
traycer host doctor
```

[...]

Issue #996, opened 2026-08-05
Could not start Traycer Host


bsec-hash:
### Environment

| Field | Value |
|---|---|
| App Version | 1.1.9 |
| Platform | win32 (x64) |
| Host | 1.1.9 (ready, pid 60164) |
| Electron | 42.7.1 |
| Chrome   | 148.0.7778.280 |
| Node.js  | 24.18.0 |
| Support Report | `rpt_dcc2a19bc42f4f898abac89f3c5cf807` |

### What happened?

Area: Host startup

Traycer Host could not start.

### Steps to reproduce

1.
2.
3.

hdkshingala:
Thanks for the report — the log bundle told the full story, and this is a known issue on our side. Two things stacked:

1. **The "Could not start Traycer Host" card was premature.** On a fresh install the startup check gives up after 60 seconds, and it fired before you had finished signing in — host setup only begins after sign-in, so nothing was actually broken at that point.

2. **After sign-in, the host started fine but immediately got buried.** Once it came up (10:07 local), it ran first-time discovery of every coding-agent CLI at once, including unpacking provider packages — some took 20+ seconds — and the host became unresponsive in bursts of up to ~30 seconds. The app kept losing and regaining its connection during that window, which looks exactly like "host won't start" from the outside. By the time your report was filed the host was actually up (the report itself captured it as `ready`).

Both parts are fixed on our development branch and will be in the next release: the premature startup card no longer fires early (#868), and first-run discovery no longer bursts hard enough to make the host unresponsive (#977, #978, plus host-side changes).

Until then it's transient: after the first couple of minutes on a fresh install the host settles and stays healthy, and restarting the app clears any lingering errors. If you run antivirus software with real-time scanning, adding an exclusion for `%USERPROFILE%\.traycer` speeds up that first-run unpacking considerably.

We'll follow up here once the release with these fixes ships.

hdkshingala:
v1.1.10 is out and contains real work on this, but I don't want to overstate it, because the specific path your report hit is **not** among it.

What shipped: host-recovery sweeps no longer re-probe every harness catalog, the stream loop counters now reset on delivered data rather than on subscribe-acknowledgement, and there is a host-side gate so the feedback loop between those two cannot re-form. That is the mechanism behind the bursts in the earlier reports on this family.

[...]

Issue #961, opened 2026-08-04
Could not start Traycer Host


Alberto1505:
### Environment

| Field | Value |
|---|---|
| App Version | 1.1.9 |
| Platform | win32 (x64) |
| Host | 1.1.9 (ready, pid 34080) |
| Electron | 42.7.1 |
| Chrome   | 148.0.7778.280 |
| Node.js  | 24.18.0 |
| Support Report | `rpt_4473cb7a42ec4dee84bacf7a7a9a38d7` |

### What happened?

Area: Host startup

Traycer Host could not start.

### Steps to reproduce

1.
2.
3.

hdkshingala:
Thanks for the report — your support bundle (and the one from #995, which is the same problem recurring) shows exactly what happened. It's two known issues stacked, same root cause as #858 and #890, and both are already fixed for the upcoming release.

**1. The first "Could not start Traycer Host" screen was a false alarm.** This was a fresh install: the app enforces a fixed 60-second startup deadline, and it fired while setup and sign-in were still in progress. The host actually started fine about three minutes later and stayed up — your report itself shows it running (`ready, pid 34080`).

**2. The real blocker: the host's sign-in verification request keeps failing.** Once running, the host needs one HTTPS request to `authn.traycer.ai` to verify your sign-in, and on your machine every attempt failed with `fetch failed` — while the app itself reached the internet fine (your browser sign-in and update checks succeeded). In 1.1.9 that network failure is mistaken for a sign-in rejection, so the app shows the error screen even though the host is alive. On Windows this selective blocking is almost always one of:

1. An **antivirus / firewall** blocking non-browser programs (or specific domains) from reaching the internet.
2. Broken or flaky **system DNS** — the app hides this by using secure DNS internally; your log also shows DNS drop-outs around laptop sleep/wake.
3. A network that requires a **proxy** — the app honors Windows proxy settings, but the 1.1.9 host does not.

**To get unblocked now:**

- If a third-party antivirus/firewall is installed, add exceptions for `%USERPROFILE%\.traycer\host\install\traycer-host.exe` and `%USERPROFILE%\.traycer\cli\bin\traycer.exe`, then restart Traycer.
- Run `nslookup authn.traycer.ai` in PowerShell — if it fails or is intermittent, switch your network adapter's DNS to `1.1.1.1` / `8.8.8.8` and restart Traycer.
- After waking the laptop from sleep, give the network a moment to reconnect before relaunching Traycer.

[...]

PR #1018, opened 2026-08-05
fix(desktop): don't wait 60s for a host that was never installed

hdkshingala:
## Problem

On a machine where **no host has ever been provisioned**, `HostLifecycle.bootstrap()` entered the readiness wait and always burned the full 60s budget before throwing `HOST_NOT_READY`.

The budget is quiet-time and re-arms only from installer progress (`notifyProvisioningActivity`), but the launch converge deliberately refuses to provision a never-installed host before sign-in (`isUnavailableInstalledHost`). So no progress event can arrive, `lastProvisioningActivityAt` stays `0`, and the deadline fires at exactly 60s every time.

The error is **log-only** — its sole subscriber calls `log.error`, it never crosses IPC, and `gui-app` has no reference to the type. It cannot render a card. But it lands at ERROR in the `desktop.log` attached to every fresh-install support report, where it has misdirected three field investigations (#961, #996, #1001) into blaming a mechanism that cannot produce the symptom under investigation. The wait also delays everything gated on `bootstrap` resolving: the host health monitor (owner of Windows auto-respawn) and the macOS login-item revision monitor.

## Change

**Gate on whether a host was ever installed, not on sign-in state.** Waiting for a host nobody installed is pointless regardless of sign-in; waiting for one that *is* installed stays correct. That collapses it to a single predicate and keeps auth state out of the lifecycle.

- `desktop-startup.ts` resolves `installedVersion !== null` at the boot seam and hands the fact to `bootstrap({ hostInstalled })`. `HostLifecycle` must not read it itself — its contract forbids consulting the service controller.
- A failed status read answers `true`, so an unreadable install record can never silently strip a legitimate readiness wait (and its `HOST_NOT_READY` signal) from a host that is installed and merely slow.
- The composition lives in `bootstrapHostWithInstallState()` rather than inline: the startup test hooks replace `runDeferredBackground` wholesale, so anything written inline there is unreachable from a test and could regress silently.

**Also: create the host root before installing the pid-metadata watcher.** `fs.watch` needs the directory to exist, and on a fresh machine it does not — so the watcher failed ENOENT and the lifecycle held none for the rest of the session (visible in all three field logs). The readiness wait had been covering this *by accident*: an install starting during the wait created the root before the watcher went in at the end of it. Skipping the wait removes that accident, so the directory is now created explicitly. The desktop already does the same `mkdir` in `writeHostNameSettings`, and nothing infers install state from the directory's existence.

## Testing

- New `host-install-state.test.ts` — install-state reads plus the boot-seam wiring (null → `false`, installed → `true`, rejection → `true`).
- New `host-lifecycle` case rooted at an **absent** nested directory, which is the real fresh-install shape, performing the CLI's actual sequence (create root, then publish `pid.json`) so the watcher is what is under test. A 30s budget against a 3s race means a regression to the waiting path fails outright rather than merely running slow.
- Both behaviours were mutation-probed: reverting either fix makes the new test fail, with a self-explaining assertion rather than an opaque timeout.
- 325 tests pass across `electron-main/startup` and `electron-main/host`; compile, lint and format clean.

## Notes

`readDesktopHostInstallRecord` is deliberately tolerant and maps read/parse failures to `null`, so a corrupt record reads as "never installed" and loses its wait. Accepted: the watcher is now installed regardless, so a host that publishes late is still picked up.


hdkshingala:
@coderabbitai approve

PR #1032, opened 2026-08-06
test(gui-app): remove redundant browser API imports (1/2)

tanveergill:
## Summary
- remove redundant per-file browser API side-effect imports
- retain the shared Vitest setup file as the single global installation point

## Validation
- pre-commit affected build, compile, lint, and format checks


tanveergill:
@coderabbitai review

PR #1079, opened 2026-08-09
fix(gui-app): accept image attachments for Kimi K3 and K2.7 models

gavasques:
## Summary

The host model catalog does not flag the Kimi K3 / K2.7 families as vision-capable, so the composer blocks image attachments for them ("Attached images are not supported by the selected model") — even though these models accept image input. The Kimi CLI's own catalog marks them as such (`kimi-code` exposes `k3` / `k3-256k` with `images=yes`).

This adds a curated client-side capability override consulted before host metadata in `modelSupportsImageAttachments()`:

- On the `kimi` harness, the generation marker (`k3` / `k2.7`) alone suffices — slugs there can be bare `k3` with label `K3`.
- Cross-harness slugs (e.g. `cline-pass/kimi-k3`, `moonshot/kimi-k3`) must also carry the `kimi` name.
- Overrides are additive only and never strip a host-claimed capability.

If the host catalog gains proper flags for these models, the override becomes a harmless no-op — happy to drop it in favor of a host-side fix.

## Related issue

None filed — can open one if you'd rather track the host-side catalog gap.

## Checklist

- [x] `bunx nx run traycer-clients-gui-app:compile` passes (typecheck); `bun run build` covered by CI
- [x] ESLint clean on the changed files; Prettier check passes
- [x] Tests added: `model-capability-overrides.test.ts` (9 tests — both match shapes + negative cases)

### DIFF
--- a/clients/gui-app/src/lib/composer/__tests__/model-capability-overrides.test.ts
+++ b/clients/gui-app/src/lib/composer/__tests__/model-capability-overrides.test.ts
@@ -1,0 +1,136 @@
+import { describe, expect, it } from "vitest";
+import type { ModelOption } from "@/components/home/data/landing-options";
+import { modelSupportsImageAttachments } from "../chat-run-settings";
+import { modelImageSupportOverride } from "../model-capability-overrides";
+
+function model(overrides: {
+  readonly harnessId: ModelOption["harnessId"];
+  readonly slug: string;
+  readonly label: string;
+  readonly metadata?: Readonly<Record<string, unknown>>;
+}): ModelOption {
+  return {
+    harnessId: overrides.harnessId,
+    slug: overrides.slug,
+    label: overrides.label,
+    description: null,
+    contextWindow: null,
+    maxOutputTokens: null,
+    defaultReasoningEffort: null,
+    supportedReasoningEfforts: [],
+    defaultServiceTier: null,
+    supportedServiceTiers: [],
+    metadata: overrides.metadata ?? {},
+  };
+}
+
+describe("modelImageSupportOverride", () => {
+  it("flags Kimi K3 slugs as vision-capable", () => {
+    expect(
+      modelImageSupportOverride({
+        harnessId: "kimi",
+        slug: "kimi-k3",
+        label: "Kimi K3",
+      }),
+    ).toBe(true);
+  });
+
+  it("flags Kimi K2.7 slugs as vision-capable", () => {
+    expect(
+      modelImageSupportOverride({
+        harnessId: "kimi",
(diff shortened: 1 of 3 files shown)


hdkshingala:
Hi @gavasques — thanks for the contribution, and especially for the clear writeup and tests. Your diagnosis of the gap is correct: the model catalog doesn't flag the Kimi families as vision-capable even though the models accept images. But the fix needs to land one layer down, and I want to explain why this PR unfortunately can't work end-to-end as-is.

**Image support is enforced in two places, and this patches only the first.** The composer check you overrode (`modelSupportsImageAttachments`) controls the attach button in the UI. But the host independently re-checks the same catalog metadata on every image-bearing send, and rejects unsupported sends with an `UNSUPPORTED_IMAGES` error ("The selected provider or model does not support image attachments."). With this override merged, the attach button would enable — and then the send would fail at the host. That's a worse experience than the disabled button: the user composes a message with images and only finds out at send time.

**The right fix is in the host's Kimi catalog, and it's small.** Kimi Code already advertises image support over ACP — the `initialize` response carries `agentCapabilities.promptCapabilities.image: true` (we verified this against kimi-code 0.34.0, the current pin). The host's send path already trusts exactly that flag when attaching image blocks to a turn; the catalog builder just never stamps it onto the model metadata the composer reads. We're fixing that on the host side: `listModels` will read the flag from the handshake and stamp `inputModalities: ["text", "image"]` / `supportsImageAttachments: true` onto the discovered Kimi models. Once that lands, the attach button enables for Kimi with no client change — and it stays live-discovered, so it tracks whatever future Kimi CLI versions advertise instead of a curated slug list.

One more reason we'd rather not take a client-side family table even temporarily: the ACP handshake is the only capability signal Kimi actually exposes (we probed `session/new` — the per-model entries carry no vision metadata), so a slug/label pattern match has nothing authoritative to anchor to, and it would need manual curation every time Moonshot ships a new generation.

You offered to drop this in favor of a host-side fix — we'd like to take you up on that and close this PR. The investigation you did here is what surfaced the catalog gap, so thank you; we'll note this PR when the host fix ships. If you'd like to file the tracking issue for the catalog gap you mentioned, we're happy to link it too.

gavasques:
Thanks @hdkshingala, clear explanation and you are right on both points. Two-layer enforcement means a client override can only ever be half a fix, and the ACP handshake flag is the only authoritative signal worth keying on.

Closing this PR in favor of the host-side catalog fix. Filed the tracking issue at #1118, happy to have it linked when the host fix ships. Appreciate the careful review.

PR #1087, opened 2026-08-10
fix(cli): keep host running after child OOM kill

tanveergill:
## Summary

- Set OOMPolicy=continue on the generated Linux systemd user unit so an OOM-killed agent or tool child does not stop the entire Traycer host unit.
- Cover unit generation and replacement of an existing unit before systemd daemon reload.

## Related issue

- None

## Validation

- Focused Linux unit, Linux install-flow, and service lifecycle tests: 35 passed.
- Pre-commit affected build, compile, lint, and format checks passed.
- DCO sign-off passed.

## Checklist

- [x] Pre-commit hooks pass
- [x] Tests added or updated where it makes sense
- [x] Commits are signed off per the DCO


tanveergill:
Addressed the replacement-test feedback in f0afa0c9: the fixture now uses a unique `stale-unit-fixture` marker, and the test proves it is absent both when `daemon-reload` observes the unit and after installation. Focused CLI tests pass: 35/35.

PR #1135, opened 2026-08-12
fix(gui-app): hide unavailable providers from Agent Teams block picker

gavasques:
## Summary
- The Block dropdown in Agent Teams listed providers marked `(unavailable)` (Amp, Copilot, Qwen Code, etc.).
- Those harnesses are already omitted from pack ladders and model selects, so blocking them did nothing.
- The picker now shows only available providers. Existing blocks still render with Unblock.

## Test plan
- [ ] Open Settings → Agent Teams → Blocked providers
- [ ] Confirm the Provider dropdown lists only signed-in / available harnesses (no `(unavailable)` suffix)
- [ ] Block an available provider or model and confirm Unblock still works
- [ ] If a previously blocked unavailable provider exists, its card still shows with Unblock


gavasques:
Opened against the upstream repo by mistake — this is a personal-fork UI tweak, not intended for traycerai/traycer.

PR #1160, opened 2026-08-13
feat(protocol,shared): whole-body mux chunking for oversized envelopes

hdkshingala:
## What

Every logical mux message now encodes to one opaque body (`[bodyFlags][jsonLen][json][binary]`) carried as 1..N ≤64 KiB frames whose binary sections concatenate — so a multi-hundred-MB `chat.subscribe` snapshot and a tiny pong ride the same code path, and the single-frame case is not a special case. The protocol is unreleased, so wire semantics change freely.

## Protocol (`@traycer/protocol`)

- New `host-transport/chunking` module: whole-body codec, pull-based `OutboundChunkSource` (frames materialize at scheduler pull time, drawing per-stream `seq`), `ChunkReassembler` (type/seq adjacency, FATAL/CLOSE mid-sequence abort-and-deliver, injectable receive cap), and a token-bucket `ChunkPacer` (6 MiB/s + 375 f/s, 1 MiB + 64-frame burst) sized so no relay fixed-window phase alignment can exceed the relay's 8 MiB / 500-frame budget.
- New `CHUNK_FIRST` mux flag marks sequence starts (frame-level `json` is always null now).
- Sender plaintext cap budgets the Noise transport overhead (25 B) **and** the relay host-leg `[sid:u32]` demux prefix — a regression seals a cap-sized frame through a real Noise session, applies host-leg framing, and pins exact equality with the relay's 1 MiB cap.
- 512 MiB logical message bound (`MuxMessageSizeError`): stringify-in-try + byte check on send, mid-accumulation enforcement on receive.
- Shallow (structural) validation schema for `chat.subscribe` snapshot histories, sound **only on the live negotiated line** (`chatSubscribeLiveSchemaVersion` exported beside the live contract).

## Client transport (`clients/shared`)

- Source-based `PriorityScheduler`: per-stream FIFO held across the interactive/bulk queues (a delta must never overtake its own snapshot's chunk tail), credit-gated bulk, pacing for chunked sources with self-rearming resume, `dropStreamOutbound` for dead streams.
- Per-frame bulk credit accounting at frame receipt, symmetric with the host's per-frame spend (per-message accounting deadlocks any transfer longer than the initial credit window).
- Per-stream failure routing: reassembly/size faults fail one stream — never the connection — with a full terminal transition (drop queued outbound, CLOSE to the host, typed local failure). A host stream FATAL now rejects a pending unary immediately instead of waiting out the 30 s timeout.
- Bounded terminal-stream tombstones (mirror of the host's frontier) drop relay-delayed chunks for dead streams before they can seed orphan reassembler accumulators.
- `chat.subscribe` snapshots take the shallow parse only when the negotiated version is exactly the live line; down-negotiated hosts keep the deep parse whose compatibility defaults up-convert pre-image messages (e.g. `imageResolutions`).

## Testing

- Shared reassembler conformance spec (11 cases) run against the protocol implementation.
- Real-Noise integration suites: chunked round-trips over the new wire, mid-transfer credit grants, concurrent-delivery ordering, per-stream fatal routing, tombstone resurrection attacks.
- Scheduler FIFO tests genuinely pace-block (>64-frame sources under fake time) and are mutation-verified against removal of the blocked-stream scan.
- 24 protocol + 38 scheduler/session + 3 chat-client tests green; the paired internal-repo PR carries the host/relay side.


hdkshingala:
@coderabbitai approve

PR #1169, opened 2026-08-14
feat(protocol,gui-app): epic.readChatAttachment - chat-plane image byte channel

tanveergill:
## What

New optional unary `epic.readChatAttachment@1.0` plus the gui-app fetch chain for chat image attachments. Part of the chat-attachments-off-the-doc train: bytes move from the epic Y.Doc's `attachments` map (whole-doc replication — a private chat's pasted image is readable by anyone with epic access) onto the owning host's chat storage plane.

## Protocol

- `(epicId, chatId, hash)` request — `chatId` is the authorization subject; the contract doc explains why a bare content address would rebuild the doc-map privacy hole behind an RPC.
- `{ok: true, bytesBase64, mediaType}` (host-authoritative magic-byte sniff, asset-stream enum reused) or `{ok: false, reason: "missing"}` — one refusal arm by design, so the response cannot become an existence oracle; transients ride the RPC error channel.
- Registered `degrade: {kind: "unsupported"}`, off the released floor. Old hosts answer `E_HOST_UNSUPPORTED`; the client falls back to today's doc-replica read.

## Client

- Chat render sites (gallery, assistant markdown images, generation cards, composer chips) resolve through a chat-scoped fetcher: blob cache → tab-host RPC → presence-guarded doc-replica fallback. `chatId` threads via a new chat-attachment scope context.
- Unsupported-host verdicts cached per host (no per-image re-probing).
- Clipboard re-inline and prompt stash fetch through the same chain with bounded timeouts.
- Artifact images stay on the doc-replica path (they're epic-shared by design).

Host/server half: internal PR (same train, lands together).

### DIFF
--- a/clients/gui-app/src/components/chat/__tests__/chat-message-user-body.test.tsx
+++ b/clients/gui-app/src/components/chat/__tests__/chat-message-user-body.test.tsx
@@ -35,6 +35,10 @@ const attachmentMocks = vi.hoisted(() => ({
     Promise.resolve(new Uint8Array([1, 2, 3])),
   ),
   hasBytes: vi.fn(() => true),
+  readChatBytes: vi.fn(
+    (_hash: string): Promise<Uint8Array<ArrayBuffer> | null> =>
+      Promise.resolve(new Uint8Array([1, 2, 3])),
+  ),
 }));
 const composerPickerMocks = vi.hoisted(() => ({
   useComposerPickerItems: vi.fn(),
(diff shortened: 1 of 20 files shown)


tanveergill:
Replaced by the re-cut branch (identical change; the original commit's Signed-off-by trailer was split from the trailer block by a blank line, failing the DCO check, and this repo's completion flow forbids force-pushes). See the new PR.

PR #1184, opened 2026-08-15
feat(gui-app): let the epic usage chart group by harness or model

tanveergill:
## Summary

The Settings usage dashboard gained a Harness/Model toggle over its daily chart in #1168; the epic usage dialog kept the same chart pinned to `"harness"` with no way to ask the other question. Same reader, same chart, two different answers depending on where it was opened. This wires the existing `UsageChartGroupByToggle` into the epic dialog.

- grouping is held in `EpicUsageDialog`, not its body: Radix unmounts `DialogContent` while closed, so a selection held in the body would silently reset on every reopen — the same reason the window picker's state lives there
- the series scale is keyed by the reader's grouping. Unlike the Settings panel, where the harness split beside the headline needs a harness-keyed scale of its own regardless of the chart, the chart is this dialog's only consumer of the scale
- the toggle stays inside the `factCount === 0` guard — with no chart to regroup, a control that changes nothing visible reads as broken
- the chart remounts on switch (`key={chartGroupBy}`) per the prop's contract, so the legend's hidden-series set cannot filter model bands with stale harness keys

## Related issue

Follow-up to #1168.

## Validation

- `epic-usage-dialog` suite: 7/7 green; React Compiler variant 10/10
- new coverage: the default harness series regroup to model series after clicking the toggle, and a no-facts window renders no toggle at all
- pre-commit affected build, compile, lint, and format checks

## Checklist

- [x] Pre-commit static checks pass
- [x] Commits are signed off per the DCO

### DIFF
--- a/clients/gui-app/src/components/epic-canvas/panels/__tests__/epic-usage-dialog.test.tsx
+++ b/clients/gui-app/src/components/epic-canvas/panels/__tests__/epic-usage-dialog.test.tsx
@@ -37,6 +37,24 @@ function chartDayCount(): number {
   return data.length;
 }
 
+/**
+ * The mounted trend chart's series names, in the last captured ECharts
+ * option's series order - reads the same "last mocked instance, last
+ * captured option" seam as {@link chartDayCount}, so a `chartGroupBy` switch
+ * (which remounts the chart via its `key` prop) is picked up by re-reading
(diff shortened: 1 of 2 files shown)


tanveergill:
@codex review

tanveergill:
@coderabbitai approve

PR #1198, opened 2026-08-15
feat(protocol,gui-app): session-scoped background stop for version-gated commands

AmiteshwarRandhawa:
## Summary

Codex promotes still-running commands to background items on every build, but the per-command stop lever (`thread/backgroundTerminals/terminate`) only exists from 0.146.0. Below that floor the panel rendered a stop button that could never work: clicking it failed with a misleading "no longer running" toast, the chat stayed permanently running, and its worktree stayed undeletable.

This is the client half - the protocol capability data and the GUI escalation path. The host half lives in the internal repo.

### Protocol (`chat.subscribe@1.7` only)

Released lines 1.4-1.6 are re-pinned to frozen pre-change copies so their surfaces stay byte-identical.

- Command background items split into their own union arm carrying `individualStopUnavailable { providerLabel, minVersion }` as data, so the renderer gates the stop button from the wire instead of learning provider versions.
- New `stopBackgroundSession` action: ends the session's provider process, taking every background item in it down together - the one stop that works on every codex version.
- `actionAck`'s action enum is parameterized in the shared frame factory and frozen for released lines.

### GUI

- Gated command rows render a disabled stop with a tooltip built from the wire data, pointing at **Stop all**.
- **Stop all** escalates to a `ConfirmDestructiveDialog` when a gated command is present (unchanged otherwise): it names the provider, the blast radius, and the active turn when one is live.
- The store sequences that escalation in two phases - stop the turn first, dispatch the session stop when the turn-settled frame arrives - because the host refuses a session stop under a live turn (a mid-turn kill reads as a crash and starts the retry loop).

## Related issue

Host half: traycerai/traycer-internal#4984 (that PR bumps the submodule pin to this branch).

## Checklist

- [x] Pre-commit static checks pass (`pre-commit run --all-files` for an explicit full-repo run)

> Note on the first box: the commit was landed with the hook bypassed. `gui-app` compile fails on a pre-existing prosemirror-view 1.42.1/1.42.2 lockfile duplication in `editor-core`, unrelated to this change. Protocol compile, the surface-compat tripwires, and every touched test suite were run green in isolation.


AmiteshwarRandhawa:
@coderabbitai review

AmiteshwarRandhawa:
@coderabbitai review

AmiteshwarRandhawa:
@coderabbitai review

AmiteshwarRandhawa:
@coderabbitai full review

AmiteshwarRandhawa:
@coderabbitai full review

AmiteshwarRandhawa:
@coderabbitai approve

PR #1216, opened 2026-08-15
fix(gui-app): clear and preserve resolved prompt glyphs

tanveergill:
GUI half of the resolved-prompt indicator work. The host half already landed in the internal repo as #4969 ("clear prompt indicators when notifications are read"), so this side had been sitting unmerged; splitting it out of an unrelated feature branch so it can ship on its own.

- Clear prompt glyphs when their notifications are read.
- Preserve resolved prompt state in the glyph projection rather than recomputing it from a stale slice.
- Clarifying comments on what "resolved" means for indicators.
- Test coverage aligned with the resolved-state behavior.

Scope is three files under `stores/notifications` and `hooks/notifications`; no protocol change.


tanveergill:
Closing: superseded and already upstream. These four commits are the same work that merged to `main` as #1180 (squash-merged, hence the different SHAs). After merging current `main` into this branch the net diff is zero files — all three touched files are byte-identical to `main`. Nothing to ship here; the branch was carrying commits that only looked unmerged by SHA. Opened in error while splitting them out of an unrelated feature branch.

PR #1230, opened 2026-08-16
fix(gui): stop the fork dialog waiting forever, and block a doomed submit

hdkshingala:
Client half of a cold review of the cross-host chat fork feature. The host half is internal #5014; this PR carries the protocol field it answers with and the two client behaviours that read it.

Follows #1227.

## `definitive` — a way to say "never"

`epic.chatPublicationState` gains `definitive`, set when waiting **cannot** change the answer. Three host states needed it and none could be expressed before:

| Reason | State |
| --- | --- |
| `chat-deleted` | the source chat is a tombstone on its own host |
| `lineage-superseded` | the chat lost an arbitrated fork; its publications land under a different cloud identity |
| `backup-halted` | publication stopped for a reason the sweep does not retry within the process lifetime |

All three previously reported byte-identically to *"the sweep has not reached it yet"*, so the condition poll re-asked every 30s forever while the dialog said **"It backs up automatically — try again shortly."** and kept submit blocked. The poll now terminates, and each reason gets copy that is true for it.

Two cases that look symmetric and are not:

- an **unrecognised** reason is treated as terminal-but-unexplained, so a newer host cannot reintroduce the infinite wait;
- an **absent** one is not. The field was added to an unreleased method in place, so a host built before it takes the un-parsed same-version path and the key genuinely arrives missing. Reading that as a reason would mark every older host permanently halted — the same hang, inverted.

## Submit no longer stays enabled on an uncovered boundary

`verdictAllowsSubmit` previously allowed submit while `boundaryCovered === false`, documented as safe because *"Layer 2 — the host's typed refusal — is the authority"*.

That premise was false. The host's check was presence-only, so a boundary turn published mid-stream is *present* in the head at its partial version: the fork returned 200 OK and seeded a **silently truncated turn**.

The host-side currency check lands alongside this in #5014. Both halves are wanted rather than either alone — the publication hook reports `UNKNOWN` during background refetches, so the gate blinks open for one RTT per poll cycle, and the host refusal is what makes that blink a wasted round trip instead of a hole.

## Preserved deliberately

- `boundaryCovered: null` is still never collapsed to `false` — `null` means the question was not answered, and blocking on it would refuse a fork on a question nobody asked.
- The permissive `unknown` path for hosts that predate the method stays permissive; an unsupported or unreachable host does not start blocking.

## Tests

84 tests across three suites, including four mutation probes — disabling the `definitive`-first check, treating `undefined` as a reason, disabling the invalidates-published-head gate, and reverting the submit block — each confirmed caught, with production restored and `diff`-verified byte-identical afterward.

### DIFF
--- a/clients/gui-app/src/components/chat/__tests__/chat-fork-dialog-cross-host.test.tsx
+++ b/clients/gui-app/src/components/chat/__tests__/chat-fork-dialog-cross-host.test.tsx
@@ -1149,7 +1149,16 @@ describe("ChatForkDialog cross-host routing", () => {
     expect(dialogMocks.createMutate).not.toHaveBeenCalled();
   });
 
-  it("boundarySyncing: notice speaks, rows stay selectable, submit stays enabled", async () => {
+  it("boundarySyncing: notice speaks, rows stay selectable, submit is blocked", () => {
+    // Clicking OTHER_HOST_ID makes this a CROSS-HOST fork, and
+    // `chatForkTargetVerdict` only reaches the `boundaryUncovered` arm (→
+    // `boundarySyncing`) when `isCrossHost` is true — a same-host fork
+    // short-circuits to `allowed` before this check ever runs (see the
+    // sibling first-paint test below, which stays same-host). `boundarySyncing`
+    // is no longer exempt from `verdictAllowsSubmit`: the host's coverage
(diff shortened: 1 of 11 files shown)


hdkshingala:
@coderabbitai approve

PR #1235, opened 2026-08-16
chore(deps): bump electron to 42.9.1 and concurrently to 10.0.5

hdkshingala:
Routine dependency refresh for the OSS workspace root, run as part of the recurring `/update-dependencies` pass.

## Summary

| dep | from | to | where |
| --- | --- | --- | --- |
| `electron` | 42.9.0 | 42.9.1 | root `catalog:` (consumed by `@traycer-clients/desktop`) |
| `concurrently` | 10.0.4 | 10.0.5 | root `devDependencies` |

Both are in-range patch updates surfaced by `bun outdated --recursive --no-progress`. The ranges were edited **by hand** and `bun install` was run afterwards, so no `catalog:` reference was materialized into a workspace `package.json` — the diff is the two range lines plus `bun.lock`.

`bun update electron` was deliberately not used: at the workspace root it resolves `electron` as a new root dependency at 43.4.0 (a major) rather than updating the catalog entry, so it was reverted in favour of the hand edit.

## Not taken in this pass

Everything else `bun outdated` reports here is an out-of-range major, left for a deliberate upgrade: `typescript` 6→7, `jsdom` 29→30, `@dagrejs/dagre` 1→3, `undici` 7→8, `encrypt-storage` 2→3, `motion` 12→13, `nx` 22→23, plus the standing `eslint` / `@eslint/js` 9.x hold (`eslint-plugin-jsx-a11y` peers do not support ESLint 10).

## Verification

`pre-commit` ran on commit and passed, including `build · compile · lint · format (nx affected)` and the DCO check.


hdkshingala:
@coderabbitai approve

PR #1247, opened 2026-08-17
refactor(protocol): rename the v7.0 pre-image schemas so the V70 names are free

hdkshingala:
Follow-up to #1239, which left this instruction in `registry.ts`:

> Whoever opens v8.0 must therefore rename that pre-image for what it is (it is a fixture, not a line) before taking the real v7.0 freeze under the `V70` names - not add a third naming scheme beside it.

This does that rename, and corrects three things that instruction got wrong.

## Why the names were occupied

`providers.list@7.0` binds the **live** schemas — the release collapsed the unreleased v8.0 into it. So the `*V70` copies of the response describe nothing v7.0 serializes: they are the pre-image the line had before the version-manager fields arrived, back when those sat on a v8.0 above it. Opening v8.0 means freezing the real v7.0, and the names it needs were taken by a different shape.

## Scope was measured, not assumed

Every `*V70` export was dumped with `z.toJSONSchema` and compared against its live counterpart. Only the ones that **differ** were renamed:

| Renamed to `*V70Preimage` | Kept bare `V70` (identical to live) |
|---|---|
| `providersListResponseSchemaV70` | `providersListRequestSchemaV70` |
| `providerCliStateSchemaV70` (+ base shape) | `providerIdSchemaV70` |
| `providerManagedInstallStateSchemaV70` | `providerLoginCapabilitySchemaV70` |
| `providerNativeCapabilitiesSchemaV70` (+ default) | `providerManagedInstallErrorReasonSchemaV70` |
| `nativeListResultSchemaV70` (+ private success arm) | `providerMcpCapabilitiesSchemaV70` |
| `providerSkillSchemaV70` | `providerPluginsCapabilitiesSchemaV70` |
| `providerSkillsCapabilitiesSchemaV70` | `providerNativeScopeSchemaV70` |
| `providerSettingsTabSchemaV70` | `providerEnvOverrideScopeSchemaV70`, MCP enums |

The criterion is what makes the right-hand column safe to leave: **a copy that still equals live is still exactly what v7.0 serializes**, so a future v8.0 author can reuse it as the real freeze. Its name blocks nothing. Only a diverged copy creates the collision. `provider-model-providers-compat.test.ts` already fails the day one of them stops matching live.

The two bridge fills were renamed with the shape they target (`upgradeProviderCliStateToV70Preimage`, `upgradeProviderCliStateListToV70Preimage`).

## Three corrections to #1239's comment

1. **It listed `providersListRequestSchemaV70` as "NOT the v7.0 wire".** Wrong — the collapse moved only the response. The frozen request is byte-identical to live and is a real freeze, so it keeps its name. Nothing enforced that equality, so this adds a pin for it (`the frozen v7.0 request is still exactly the live request`), verified to fail by adding a field to the live request.
2. **It said the pre-image "survives only as a fixture".** Wrong — the v6 → v7 bridge re-parses *every* upgraded row through it in production (`upgradeProviderCliStateListToV70Preimage`, `registry.ts`). The comment now says not to delete it, and why.
3. **It said `rate-limit/schemas.ts` parses through it.** Stale — #1239 itself deleted that binding.

## Orphaned projection deleted

`projectNativeCapabilitiesToV70Preimage` and its private skills helper had **no production caller**: the v8→v7 hop they fed was deleted by the collapse, leaving only their own test. Both are removed (second commit).

The three cuts they encoded were decisions rather than code, so those stay written down where the functions were — filter `supportedTabs` rather than reparse it (one unknown member fails the whole array, and the `.catch()` then costs the reader MCP, Plugins and Skills over a single tab id), destructure `modelProviders` out, drop the Skills `inspect`/`edit`/`update` actionScopes. The two comments that pointed at the projection now carry that instruction instead of a dangling reference.

The tests pinning the projection's *premise* stay: frozen tabs = live tabs minus `modelProviders`, frozen capability keys = live keys minus `modelProviders`. Only the test driving the deleted function goes.

## Verification

- Protocol surface dump **byte-identical to `origin/main`** at 9,734,342 bytes — no wire shape moves.
- `bun run compile` clean; protocol host suites **642/642 pass** (23 files).
- eslint `--max-warnings 0` clean on the changed files.
- No consumer outside `protocol/` referenced any renamed or deleted symbol (checked the host repo and `clients/`).


hdkshingala:
@coderabbitai approve

Issue #17, opened 2025-08-03
"Database locked" when using in multiple VS code instances


dfcfdev-partners:
## Describe the Issue

I tend to run up to 4x projects in parallel. No abuse, i just have to get a lot of work done on multiple webdev customers projects.

When i try to run a second or a third instance in planning phase in parallel i sometimes get the message "Failed to generate phase: Database locked ... [details i forgot to copy when this occured]"

I would very much like to be able to run up to 4-6x projects in parallel.

I have tried Cursor, Claude Code, Kiro and every AI agent and mcps or setups for planned thinking. Traycer just gets it right and it would be wonderful if you could make it possible to run more instances in parallel.

## Type of Issue

- [ ] Performance
- [ ] Security
- [ x ] Usability
- [ ] Other

## Language

- [ x ] JavaScript
- [ x ] TypeScript
- [ x ] Python
- [ ] Golang

## Steps to Reproduce

Steps to reproduce the issue, if possible:
1. Xubuntu 22.04
2. 2-4 instances of vs code with Traycer installed
3. Try to run phase discussion on 2-3 instances of VS Code / Traycer

## Expected Behavior

I would like to be able to chat with Traycer and make it plane 4-6 projects or more in parallel.

## Actual Behavior

When trying to run 2-3 Traycer instances in parallel (starting with 9/9 artifacts or recharging 0.50$ on demand) it often fails with the Database locked error described above.

## Version Information

- Extension version: (e.g., "v0.1.0") newest
- VS Code version: (e.g., "1.89.1") newest
- OS version: (e.g., "Ubuntu 20.04") Xubuntu 22.04

hdkshingala:
Hello @dfcfdev-partners,

Thank you for bringing this to our notice.
We did some fixes around this in release v2.9.9. So, please update the extension if you haven't already and let us know if the issue persists.

We will also replicate the issue you observed with parallel VSCode instances.

PR #1248, opened 2026-08-17
fix(gui-app): retain recently-active pane chats instead of remounting them

tanveergill:
## Problem

Switching tabs inside a canvas pane — or closing a tab that covered a chat — made the chat underneath visibly re-converge its scroll position for a couple of seconds. Switching **top-level task tabs** never did, and that asymmetry was the tell: a top-level switch only *hides* the surface, while an inner switch fully **unmounted** the chat, so the remount restored through LegendList's estimate-driven `initialScrollIndex` and then settled via bounded measured reissues, in full view of the reader.

Chat tabs were excluded from every keep-alive path (decision #17), and hosted-surface membership admitted only each pane's *selected* chat.

## Fix

Panes now retain their two most recently-active chats, so an inner switch becomes the same hide/show a backgrounded top-level tab already took. **No new restoration logic** — `chat-tile.tsx` already derives `surfaceVisible = paneVisible && tabSelected`, so a retained-but-deselected chat takes the existing path and `ChatMessages` replays its saved anchor once against a list it never stopped measuring.

Retention is derived from `TilePane.activationHistory` (already MRU-ordered, pruned to live tabs, persisted with the canvas).

### The subtle part

Two layers consume that policy — membership (outside React, from a store snapshot) and the pane's rendered slot set. They must not disagree: a member whose slot stops rendering keeps a non-null environment plus the sticky `canMountBody` latch, stranding its body mounted on a disconnected anchor, invisible, holding a chat session lease.

Sharing the policy function is **not** sufficient. The window is capped and selection keeps filling on rejection, so two callers passing different predicates produce two *shifted* windows rather than a subset and a superset. So the window is picked on **tile kind alone**, with caller-specific eligibility applied to the result **after** the cap.

A record's paint and geometry decisions now key off `topLevelVisible && tabSelected`: every record of a pane shares that pane's rect, so a retained one must neither paint over the selected chat nor reflow at zero width. Retained chats conceal with `visibility` rather than `display:none` for the same reason.

## Tradeoff, recorded in the code

Retention deliberately does *not* collapse while a pane is hidden — collapsing would reintroduce the churn on the first inner switch after every return from a background tab. Cost: a mounted chat leases a session that sits outside the warm-session cap and never idle-expires, so a 5-tab × 3-pane workspace roughly doubles its concurrent `chat.subscribe` sockets.

## Testing

- `bun run compile` clean; `bun run lint` clean at `--max-warnings 0`; 4,709 tests pass across the canvas/chat/store/tabs/epic-tabs suites.
- Decision-#17 tests were **converted, not dropped** — including the lifecycle-matrix negative control (re-pointed at retention eviction, where a real remount still happens) and the internal-tab-bottom-follow regression, whose teeth depend on an actual remount.
- Every new invariant was ablation-verified: reverting retention fails 5 suites; reverting the paint predicate fails the new matrix row `11b`, which drives the real store → slot → publish path rather than hand-writing `tabSelected`.

## Not done

The decision-#32 live dev-desktop gate is **only partially re-recorded**. Verified live on Electron 42.9.1: no remount across an A→B→A round trip, the retained-unpainted contract, and geometry retention. Not yet exercised: scroll-churn and hit-testing (the dev slot's chats fall back to published copies, so the assertions were vacuous), focus/IME, DnD, and the divider-drag re-drive.


tanveergill:
@coderabbitai approve

PR #1249, opened 2026-08-17
fix(gui-app): keep unanswered-question glyphs lit, and clear notifications in both modes

tanveergill:
## Summary

An unanswered question or approval kept its amber glyph only while the notification was unread. Focusing the chat marked the row read, so the glyph vanished and the running spinner took over — even though the question was still waiting.

`indicatorContribution` now keys `pendingApproval` / `pendingInterview` on `resolvedAt` alone. The bell and Needs attention list stay read-based. After that split:

- **Clear notifications** is available in both local and cloud feed modes, with the same copy. Local calls `host.notifications.clearAll`.
- Chat-scoped rows roll into their task tab in cloud mode, matching host SQL after `liveChatClause` is removed.

## Related issue

Paired with the host/server PR in traycer-internal (`traycer/prompt-glyph-resolution-state`).

## Checklist

- [x] Pre-commit static checks pass (hook ran on commit)


tanveergill:
@coderabbitai approve

PR #1250, opened 2026-08-17
docs(gui-app): correct the decision-#17 premise left behind by pane chat retention

tanveergill:
Follow-up to #1248 (cold-review findings; text only, no behavior change).

Retaining recently-active pane chats reversed **decision #17**, but ten comments still asserted the old premise as fact.

**Four were load-bearing.** The A2A store, the activity-group store, and the tile-find store each justify their module-scope, instance-keyed registries with *"chat tiles fully remount per tab switch (decision #17)"*. Nothing was broken — they key by instance id — but the stated reason no longer matched the code, so the next person deciding whether that machinery is still needed would reason from a premise #1248 deleted. It **is** still needed: a chat tile still fully remounts on retention-cap eviction, a close, or a reopen. Each comment now says that instead.

**Two were in #1248's own blast radius.** `stable-tile-surface-host.tsx` still described the pre-change `topLevelVisible`-only visibility predicate rather than `isTileSurfacePresented`, and `stable-tile-surface-host-switch.ts` still called a tab switch "the only intentional remount".

**One test reasserted a retracted claim.** The hidden-pane retention test said membership "cannot observe" pane visibility. #1248's module doc already corrected that to a *tradeoff* — `computeRetainedTopLevelRefKeys` derives the same active-ref set `TopLevelTabHost` turns into `usePaneVisible()`. That test is the one pinning the non-collapse decision, so it's the file someone opens to revisit it; it now states the real reason and the cost.

**One comment was mispriced.** Eligibility-after-cap was described as costing "one concealed placeholder" — that's the render side only. The membership side consumes a retention slot, so a pane holding a dead chat among its two most recent retains one live chat and the second still churns. Loss of benefit, not a defect, but worth saying.

**One nit:** `isRetainedChatSurface` was a one-line synonym for `isRetainablePaneChat`. The design requires exactly one name for that predicate; collapsed.

## Verification

`bun run compile` clean, `bun run lint` clean at `--max-warnings 0`, 304 tests pass across the retention, surface-host, chat-messages and tile-find suites.


tanveergill:
@coderabbitai approve

PR #1252, opened 2026-08-17
fix(protocol): chat-sync preservation — null reader floor, publisher-derived invariant, drop the dead clone seed

tanveergill:
Protocol half of the chat field-preservation train (companion host PR in the internal repo pins this branch).

## What

- **Null reader floor**: the chat-head writer schema no longer pins \`minReaderVersion\` to the 1.1 floor — it inherits nullable/default-null. The 1.1 schema additions are additive and read-safe for a 1.0 reader, and per \`COMPATIBILITY.md\` preservation is never a reason to raise the floor. \`CHAT_SYNC_1_1_READER_FLOOR\` is kept (also the cut-plan completeness threshold) and re-documented as the mechanism for a future *deliberate* raise.
- **Publisher-derived invariant**: \`COMPATIBILITY.md\` §6/§7 and the captured-levels manifest now state that parts, \`cdc\`, and \`hostPrivateShard\` are publisher-derived levels that must never carry durable chat data — re-derivable from the owner's op log, with the only round-trip through the owner's own predecessor head.
- **Drop the dead clone seed**: \`clients/shared/cloud-chat/clone.ts\` (\`buildChatCloneSeed\` / \`chatCloneResidualsOf\`) had no production consumer; the host's chat-sync record adapter is the surviving exhaustive cross-boundary consumer of the captured-levels manifest, now referenced by role.

## Verification

- \`bun run --filter @traycer/protocol test\`: 162 files / 2594 tests green
- \`bun run --filter @traycer-clients/shared test\`: 65 files / 1035 tests green
- Writer-floor mechanisms ablation-verified in the host suites (companion PR)


hdkshingala:
@coderabbitai approve

PR #1254, opened 2026-08-17
fix(gui-app): tolerate an assistant record that never carried imageResolutions

hdkshingala:
A chat tile could crash to its error boundary with `Invalid value used as weak map key`. Reported from the dev desktop app; the fix is lifted out of the host-lifecycle branch (#1243) so it can land on its own rather than waiting on that epic.

## The bug

A snapshot delivered on the live schema line takes `ChatStreamClient`'s **shallow** parse path, which validates `chat.messages` with `z.custom<Message>(isStructuralRecord)`. That is a structural check only — no zod defaults run, so `imageResolutions: z.array(...).default([])` never fills. A host replaying a record persisted before the field existed hands it straight through: typed as present, genuinely `undefined`.

`rendered-messages.ts` then keys a WeakMap on that value to build its memo signature, and `WeakMap.set(undefined, …)` throws. The whole chat tile comes down through its error boundary.

## The fix

Route the three persisted-record reads through `assistantImageResolutions()`, which falls back to one shared module-level empty array.

Sharing that array is load-bearing rather than tidiness: `imageResolutionsIdentityToken` keys a WeakMap on the value to build the memo signature, so a fresh `[]` per read would mint a new identity on every projection and defeat the cache — turning a visible crash into a silent re-render loop. There is a dedicated test for that property.

The live (`LiveAssistantMessage`) reads are deliberately left alone: that type is built in the renderer and its array is always real.

## Why here and not in the transport

An earlier attempt patched `ChatStreamClient` instead and had to rewrite a contract test to fit, which is backwards. The shallow path is structural **by design** — that is what makes it cheap — and `chat-stream-client.test.ts` pins it to hand a live snapshot's message through structurally unchanged. That test is untouched here; the tolerance belongs in the consumer that cannot accept the value.

## Verification

Two regression tests, both checked against the unfixed source — each fails with the exact `TypeError: Invalid value used as weak map key` this repairs, so they guard the behaviour rather than merely passing.

- `rendered-messages-assistant-images.test.tsx`: 14/14 pass
- `traycer-clients-gui-app:compile`: "Successfully ran target compile" (nx cache skipped)

### DIFF
--- a/clients/gui-app/src/stores/chats/__tests__/rendered-messages-assistant-images.test.tsx
+++ b/clients/gui-app/src/stores/chats/__tests__/rendered-messages-assistant-images.test.tsx
@@ -722,3 +722,86 @@ describe("useRenderedMessages image resolution stable-row identity", () => {
     ).toBeUndefined();
   });
 });
+
+/**
+ * An assistant record persisted before `imageResolutions` existed, shaped the
+ * way one actually reaches this projection at runtime.
+ *
+ * A snapshot delivered on the live schema line takes `ChatStreamClient`'s
+ * SHALLOW parse path, which validates `chat.messages` with
+ * `z.custom<Message>(isStructuralRecord)` - a structural check that runs none of
+ * the zod defaults, so `imageResolutions: z.array(...).default([])` never fills.
+ * (The DEEP path does fill it, which is why only the live line is exposed; both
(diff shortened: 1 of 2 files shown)


hdkshingala:
@coderabbitai approve

PR #1260, opened 2026-08-17
fix(gui-app): harden diff-edit browser regression startup

tanveergill:
## Summary
- Launch Chrome with `--remote-debugging-port=0` and parse its own DevTools URL instead of reserving a port ahead of time.
- Increase the DevTools startup deadline and drop the unused DBus session address in headless CI.
- Surface Vite overlay errors in browser regression timeouts for easier diagnosis.

## Test plan
- [x] `bun run test:diff-edit-browser`


tanveergill:
@coderabbitai approve

PR #1265, opened 2026-08-17
feat(protocol,gui-app): Cursor account rate-limit indicator

tanveergill:
Adds Cursor to the provider rate-limit rail, alongside codex/claude-code/grok/opencode and the credit providers.

## Why this reads a private API

Cursor's API-key-authenticated **public** API carries no account quota. Probed directly, it exposes only `/v1/me`, `/v1/models`, `/v1/repositories`, `/v1/agents`, and a **per-agent** token/cost read; `@cursor/sdk`'s `getUsage()` is scoped to one agent's runs, and the SDK reads only `retry-after` on a 429 — reactive, never forward-looking. There was no data source for a windowed rail.

The account numbers live on `aiserver.v1.DashboardService/GetCurrentPeriodUsage` (`api2.cursor.sh`). Notably `cursor-agent` bundles that method's proto but **never calls it**, so there was no existing call site to copy.

## Credential: exchange, not keychain

The host mints a dashboard session from the **configured API key**:

```
POST /auth/exchange_user_api_key            Bearer <apiKey>      -> { accessToken, refreshToken }
POST DashboardService/GetCurrentPeriodUsage Bearer <accessToken> -> current-period usage
```

Deriving the session *from* the billing key is load-bearing: it makes the account this arm reports and the account Traycer bills the same account **by construction**. The alternative — reading Cursor's own CLI session from the OS keychain — reports whichever account last ran `cursor-agent login`, needs a hand-written identity gate that can drift, and needs WorkOS refresh rotation (the stored token is routinely expired).

## The `totalPercentUsed` trap

The payload offers a ready-made `totalPercentUsed: 11.5384`, and it is **wrong for this purpose**. On the same payload `totalSpend/limit` = 28846/40000 = **72.1%**, and Cursor's own `displayMessage` reads *"You've used 72% of your included usage."* Using the pre-computed field would put a number in the rail contradicting the sentence beside it. `usedPercent` is therefore derived, and a regression test pins it against both 72.115 and not-11.5384.

Two further payload properties the parser respects:
- int64 cycle bounds arrive as **JSON strings**, guarded so a malformed bound can't become `NaN` in a countdown;
- **proto3 JSON omits zero-valued fields**, so absent is indistinguishable from zero — every payload-derived field is nullable, and a missing plan limit yields a `null` window rather than a synthetic 0% (which would render a *healthy* bar for an account whose usage nobody knows).

## Shape and wire

Hybrid like grok's — a synthesized `cycle` window feeding the shared severity rollup plus raw money — **not** credit-shaped like openrouter/kilocode/huggingface, since Cursor's money backs a real billing-cycle percentage. It therefore needs no special-casing in `classifyProviderRateLimits`, and is deliberately absent from the credit-provider severity exception in `profile-usage-projection`.

The arm rides the still-unreleased **4.0** line rather than opening a 5.0: the released baseline tops out at major 3 for both `host.getRateLimitUsage` and `agent.getProviderProfileRateLimits`, so no peer has ever negotiated 4 — the same reasoning that put Hugging Face and OpenCode there. `mapCursorAvailableToUnavailable` degrades the arm on every bridge below 4.0; because `"cursor"` is in every frozen provider-id enum (it predates Hermes/omp), the degraded row reparses cleanly down to v1.2, which also lets the three a2a profile bridges **degrade** instead of failing closed with `DOWNGRADE_UNSUPPORTED`.

Units differ within one payload: `planUsage` money is integer **cents** (divided to USD at the host boundary), while spend-limit fields are whole **dollars**.

## Verification

Endpoint, auth scheme, and response shape confirmed **end to end against a live account** — not inferred from a reference implementation. It remains a private API with no compatibility promise, so parsing is narrow and defensive: an unrecognized 200 degrades each field to `null`, total failure degrades to an explicit unavailable reason. An upstream shape change makes the rail honest, never wrong.

- protocol: 163 files / 2644 tests green
- gui-app rate-limit surface: 16 files / 210 tests green
- new coverage: cursor arm + all four downgrade bridges

The host-side fetcher and dispatch land separately in the internal repo, which will re-pin the gitlink once this merges.

### DIFF
--- a/clients/gui-app/src/components/settings/panels/__tests__/provider-rate-limit-views.test.tsx
+++ b/clients/gui-app/src/components/settings/panels/__tests__/provider-rate-limit-views.test.tsx
@@ -1262,6 +1262,7 @@ describe("CursorRateLimitView", () => {
     includedLimitUsd: 400,
     usedUsd: 325.37,
     remainingUsd: 74.63,
+    bonusUsedUsd: null,
     onDemandLimitType: "user",
     onDemandLimitUsd: 1,
     onDemandUsedUsd: 0.25,
@@ -1304,6 +1305,34 @@ describe("CursorRateLimitView", () => {
     expect(screen.getByText("On-demand limit")).toBeTruthy();
     expect(screen.getByText("$1.00")).toBeTruthy();
     expect(screen.queryByText("$100.00")).toBeNull();
+  });
+
+  it("shows overflow honestly once spend runs past the purchased allowance", () => {
+    // Past $400 the account is on Cursor's bonus grant - nothing is limited
+    // and nothing is billed, so the meter must NOT dress this up as a limit
+    // event: fill pins at 100% in the amber running-low tone (red stays
+    // reserved for the bucket bars, the actual gates), the detail keeps the
+    // REAL spend, "left" clamps at $0.00 instead of going negative, and the
+    // bonus spend gets its own row.
+    const { container } = render(
+      <CursorRateLimitView
+        data={{
+          ...cursor,
+          usedUsd: 412.1,
+          remainingUsd: -12.1,
+          bonusUsedUsd: 12.1,
+        }}
+        variant="popover-overview"
+      />,
+    );
+    expect(screen.getByText("$412.10 / $400.00")).toBeTruthy();
+    expect(screen.getByText("Included usage left")).toBeTruthy();
+    expect(screen.getByText("$0.00")).toBeTruthy();
+    expect(screen.queryByText("-$12.10")).toBeNull();
+    expect(screen.getByText("Bonus usage")).toBeTruthy();
+    expect(screen.getByText("$12.10")).toBeTruthy();
+    expect(container.querySelector(".bg-amber-500")).toBeTruthy();
+    expect(container.querySelector(".bg-red-500")).toBeNull();
   });
 
   it("falls back to the billing-cycle range when no bucket was reported", () => {
--- a/clients/gui-app/src/components/settings/panels/__tests__/provider-rate-limit-views.test.tsx
+++ b/clients/gui-app/src/components/settings/panels/__tests__/provider-rate-limit-views.test.tsx
@@ -14,6 +14,7 @@ import { RunnerHostContext } from "@/providers/runner-host-context";
 import {
   ClaudeRateLimitView,
   CodexRateLimitView,
+  CursorRateLimitView,
   GrokRateLimitView,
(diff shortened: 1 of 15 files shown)


tanveergill:
Verified and fixed in 8bd7c868 — this was real.

`projectedLiveWindows` returns `[]` for a `cycle: null` cursor snapshot, so it reached `emptyDetailProjection`; `classifyProviderRateLimits` answers `unknown` (not `limited`), so it fell to the final branch and, with `providerRateLimitWindows(...).length === 0`, reported `unavailable`/`missing_windows`. A reachable account that simply reported no allowance rendered as a fetch/account failure.

This shape is reachable rather than theoretical: the cycle window is synthesized only when the payload carries a plan limit, and the endpoint speaks proto3 JSON, which omits zero-valued fields — so `limit` can be absent rather than `0`.

Extended the existing unmeasured arm exactly as suggested, keeping expired handling untouched.

Two regression tests were added:
- `projects a cycle-less Cursor snapshot as unmeasured, not unavailable` — pins `not_checked` / `severity: unknown`.
- `still meters a Cursor snapshot that reports a cycle window` — guards the new arm from over-reaching, so a measured cycle keeps its percentage (`kind: "detail"`, 72%).

Ablation-verified: removing just the `cursor` clause turns the first test red, so it fails for the intended reason. Suite green at 18/18 in that file, and 156 across the rate-limit surface.

tanveergill:
@coderabbitai review

tanveergill:
Verified and fixed in a708a5a5 — real, with a house precedent.

The money fields are independently nullable by contract (proto3 JSON omits zero-valued fields), so a payload can carry `totalSpend` without `limit`; gating the whole line on `includedLimitUsd` hid known spend. Fixed by mirroring the Hugging Face arm's spend-only fallback in this same file (line ~148): full `remaining/limit` line when the denominator exists, `spend: X this cycle` (plus remaining when present) when it does not, and nothing only when nothing is known.

Deviated from the committable suggestion in one detail: its limit-null arm rendered `usage: unknown remaining (X used)` when `remainingUsd` is null — which is the overwhelmingly common shape, since Cursor computes `remaining = limit − totalSpend` and can't send one without the other. The spend-only wording avoids printing "unknown" for a field that's structurally absent rather than unknown.

Regression test added: null limit + present spend renders `spend: 325.37 this cycle`, no `included usage:` line, and no "unknown remaining". 13/13 in the file.

tanveergill:
@coderabbitai review

PR #1272, opened 2026-08-18
feat(gui-app): cycle History search matches with the arrow keys

tanveergill:
## Summary

Typing in the History search box narrowed the list but left no way to reach it from the keyboard — the caret stayed in the input and the only route to a match was the mouse. **ArrowDown** now steps out of the query into the results, **ArrowUp/ArrowDown** walk them, and **ArrowUp off the first row** lands back in the query.

Sibling to #1270 (keyboard traversal of the notifications feed), for the History list.

## Traversal

Row order is read fresh from the DOM on every keystroke rather than held as an index, so a refetch or a refined query that reshuffles the list mid-cycle cannot leave the cursor pointing at a row that moved, and the rows "Show more" appends join the sequence with no extra bookkeeping.

**Real DOM focus roves, rather than `aria-activedescendant`.** History rows are rich cards (link, checkbox, pin, delete, context menu), not listbox options — so `Enter` activates natively on the anchor, and the existing `group-focus-within/list-row` reveals (pin, PR pills) plus the control's own focus ring light up with no new visual state to maintain.

The handler binds to each row's full-card activation control — the `<Link>` overlay normally, the toggle `<button>` in selection mode — rather than to the list container: a keydown listener on a non-interactive `<ul>` is exactly what `jsx-a11y/no-noninteractive-element-interactions` rejects in app code, and those overlays have no focusable children to intercept.

Two deliberate edges:

- ArrowDown is only claimed (`preventDefault`) once there is a row to move to, so an empty result set leaves the caret's own behaviour intact.
- ArrowDown on the last row holds rather than wrapping, matching the `epic-sidebar-artifact-search` combobox already in the codebase.

Lands for the History modal and the `/epics` route (both `variant="page"`) plus the destination picker's toolbar search.

## Test plan

- [x] `bun run test src/components/epics/__tests__/epics-list-panel.test.tsx` — 46 passed, across both the plain and React-Compiler vitest projects
- [x] `bun run compile` for the gui-app workspace
- [x] ESLint clean on all three files (`--max-warnings 0`)
- [x] Ablation-verified: removing the row keydown binding turns the new cycle test red; restoring it turns it green

Two new tests — the full down/down/up/up-to-input cycle, and a guard that an empty result set does not swallow the key.

### DIFF
--- a/clients/gui-app/src/components/epics/__tests__/epics-list-panel.test.tsx
+++ b/clients/gui-app/src/components/epics/__tests__/epics-list-panel.test.tsx
@@ -17,6 +17,7 @@ import {
 } from "@tanstack/react-router";
 import {
   cleanup,
+  createEvent,
   fireEvent,
   render,
   screen,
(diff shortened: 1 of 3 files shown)


tanveergill:
@coderabbitai approve

PR #1288, opened 2026-08-19
fix(gui-app): open the minimap hover at the row it says is current

tanveergill:
## Summary

Hovering the minimap rail opens the list card scrolled to the wrong place.
Sitting at the end of a long chat, the card opens several rows short of the
end instead of on the turn the rail is highlighting.

**Why.** The card measured itself with `getBoundingClientRect()` inside the
layout effect that runs on the mount that also starts its own `zoom-in-95`
entry animation. That animation's `from` keyframe is live on the first style
resolution — the very one the rect read forces — so every measured distance
came back scaled by 0.95, and the reveal under-scrolled by 5% of the distance
it needed. Deep in a long list that is several rows.

A second, smaller error rode along: the reveal aligned the row's bottom edge
to the scroller's bottom edge, so the container's own bottom padding stayed
out of view and the last row sat flush against the card's edge rather than at
the end of the track.

**How.** Measure in layout coordinates instead — `offsetTop` / `offsetHeight`
/ `clientHeight` / `scrollHeight` / `scrollTop`, none of which a transform can
skew — and count the scroller's padding as part of an end row, so the first
row lands at 0 and the last at the true end. The scroller carries `relative`
so it is the rows' offsetParent and `offsetTop` is read in its own content
coordinates.

The math moves into `resolveMinimapListScrollTop` in
`minimap-track-geometry.ts`, beside the rest of the shared rail geometry, and
stays DOM-free. Both rails that share this card — the chat turn rail and the
artifact heading rail — are fixed by it.

## Tests

`minimap-list-card.test.tsx` renders the real card over a faked layout and
deliberately leaves `getBoundingClientRect()` at jsdom's all-zero default, so
restoring rect measurement turns the "very end" and "mid-list" cases red
(verified by ablation: 2 failed / 2 passed, then restored to green). Seven
geometry cases in `minimap-track-geometry.test.ts` cover the ends, mid-list,
an already-visible row, a row taller than the viewport, and non-finite input.

Locally green before the commit: `bun run compile`, `eslint --max-warnings 0`,
`prettier` (unchanged), and the minimap + chat-rail + artifact-rail suites
(42 + 10 passed). The pre-commit hook's `build · compile · lint · format`
(nx affected) and DCO gates passed on commit.

Caveat worth stating for review: the diagnosis is from the code path, not a
live browser repro. The mechanism is deterministic (the entry animation's
`from` keyframe applies on the style resolution the rect read forces), and the
padding error is independent of it and fixed too.

## Related issue

None.

## Checklist

- [x] Pre-commit static checks pass (`pre-commit run --all-files` for an explicit full-repo run)

### DIFF
--- a/clients/gui-app/src/components/minimap/__tests__/minimap-list-card.test.tsx
+++ b/clients/gui-app/src/components/minimap/__tests__/minimap-list-card.test.tsx
@@ -1,0 +1,156 @@
+import { cleanup, render } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  MinimapListCard,
+  type MinimapListEntry,
+} from "@/components/minimap/minimap-list-card";
+
+/**
+ * jsdom has no layout, so the card's geometry is faked here: rows are a fixed
+ * height in DOM order, and the scroller reports a viewport shorter than its
+ * content. The fake is installed on the LAYOUT properties only (`offsetTop` /
+ * `offsetHeight` / `clientHeight` / `scrollHeight`) and leaves
+ * `getBoundingClientRect()` at jsdom's all-zero default — measuring through
+ * rects is the defect this suite pins, because the card mounts inside its own
+ * `zoom-in-95` entry animation, where a scaled rect reads every distance short.
+ */
+const ROW_HEIGHT = 28;
+const VIEW_HEIGHT = 160;
+const ITEM_COUNT = 20;
+const CONTENT_HEIGHT = ROW_HEIGHT * ITEM_COUNT;
+const MAX_SCROLL = CONTENT_HEIGHT - VIEW_HEIGHT;
+
+const ITEMS: ReadonlyArray<MinimapListEntry> = Array.from(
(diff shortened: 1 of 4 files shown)


tanveergill:
Superseded by #1286, which landed the same fix an hour before this branch was
cut — same diagnosis (the `zoom-in-95` entry animation is live on the style
resolution the measurement forces), and a better-shaped fix: delegate to
`scrollIntoView({ block: "nearest" })` so the browser maps into the scroller's
own coordinate space and the ancestor scale factors out by construction,
rather than working around it with hand-rolled rect math.

Closing unmerged. One residual this branch also covered, recorded here in case
anyone wants it later and not as a request: `block: "nearest"` aligns the row
to the scrollport edge, so the scroller's own `pb-2` stays below the fold and
the last row lands flush against the card's edge instead of at the true end of
the track. If that ever looks wrong, `scroll-mt-1 scroll-mb-2` on the rows
fixes it — `scrollIntoView` honours `scroll-margin` — with no JS change.

PR #1303, opened 2026-08-20
feat(history): filter tasks by the hosts that own their chats

tanveergill:
Adds a **Hosts** section to the task-history filter popover, alongside Ownership / Repositories / Workspaces. It filters by the hosts that own **your own chats** in a task (`chats.owner_host_id`) — agents live on a host, and a task acquires one when you create a chat there. Deliberately not the workspace association, which is bound at epic-create and never follows the work.

The cloud half is in `traycer-internal`.

## Scope: your own chats, not every chat you can read

The dimension is scoped to the caller's own live chats. A collaborator's host is a machine you cannot name — your host directory has never seen it, so the picker would render a raw id — and surfacing which machine a teammate works on is an association nothing else in the product exposes.

It also settles duplicate chat ids rather than working around them. Identity is `(task_id, owner_user_id, id)`, so two users can hold the same chat id in one task, and the real readers resolve exactly one (`viewerPrecedence`: your row wins, else the oldest). Aggregating over every task-visible chat would ignore that precedence and report a host for a row the canonical reader hides. Within one owner a chat id is unique by the primary key, so the ambiguity stops being reachable.

**Behaviour worth noting:** on a shared task where only a teammate has chats on host X, filtering by host X will not match it. That is the intended trade for a Hosts list containing only machines you can name.

## Protocol

`epic.listTasks@1.3` carries the `chatHostIds` filter and a `chatHosts` facet group. `epic.getTaskContexts@1.2` carries the same per-row `chatHostIds` the list rows now do.

The pre-1.3 request/response/row shapes are frozen and **shared by every older minor**. Pointing one of them at the canonical latest schema instead makes the registry validator read the new field as one an older minor *drops* — which throws at registry import, not at type-check. That is how it was caught. A second instance of the same class (a response that re-pointed `facets` but not `tasks`, silently stripping `chatHostIds` from every row) was caught only by the server's compile, so there is now a test that *parses* rather than typechecks.

## Both skew directions fail closed

Two paths would otherwise render an unfiltered list as a filtered one, and in both the response comes back well-formed — there is no error to notice:

- a host below `@1.3` parses the request through its older schema, which silently strips `chatHostIds`;
- the cloud tier has no version negotiation at all, so an old server simply drops request keys it does not recognize.

The gate is the **negotiated minor**, not method presence: `epic.listTasks` is a released-floor method, so "does this host have it" is always yes and answers the wrong question. The cloud arm is a first page that came back without the `chatHosts` group — including a response carrying no `facets` object at all, since this query never sends a cursor and its response is always a first page. Either way rows are **withheld** behind their own empty state.

## Rows carry their own chat hosts

`ListTaskLight.chatHostIds` lets the filter be re-applied client-side in the two places the server's filter never reached: **id-fetched rows** (branch / worktree-path / PR-number search only matches local worktree metadata, so those tasks bypass the list filter) and **cached rows** mid-request.

A row whose `chatHostIds` is `null` is **excluded** while a host is selected — nothing has checked it against the filter, and both untrusted sources above can produce one. `[]` is different: a truthful "none of my chats on any host". A peer too old to report the field never reaches the predicate; the version gate withholds the whole list first.

`chatHosts` is also carried through the optimistic cache patchers. Its absence is the sentinel the gate reads, so a delete or create that rebuilt the facets without it would strand the list behind "Can't filter by host here" — permanently, since those entries cache with `staleTime`/`gcTime` at `Infinity`. The create path additionally declines its optimistic insert while a host filter is active, because a create projection cannot evaluate `chatHostIds`.

## Tests

Protocol suites green. GUI coverage for the two fail-closed paths, host + branch-search, URL round-trip, `any` vs `all`, the `null`-exclusion rule, and the facet-group sentinel surviving a delete.

### DIFF
--- a/clients/gui-app/src/components/epics/__tests__/epics-list-panel-phase-activation.test.tsx
+++ b/clients/gui-app/src/components/epics/__tests__/epics-list-panel-phase-activation.test.tsx
@@ -171,6 +171,7 @@ function phaseHistoryItem(): HistoryItem {
     updatedBucket: "today",
     linkedRepos: [],
     linkedWorkspaces: [],
+    chatHostIds: null,
     pullRequestNumbers: [],
     worktreeBranches: [],
     worktreePaths: [],
--- a/clients/gui-app/src/components/epics/__tests__/epics-list-panel.test.tsx
+++ b/clients/gui-app/src/components/epics/__tests__/epics-list-panel.test.tsx
@@ -269,6 +269,7 @@ function historyItem(overrides: Partial<HistoryItem>): HistoryItem {
     updatedBucket: "today",
     linkedRepos: [],
     linkedWorkspaces: [],
+    chatHostIds: null,
     pullRequestNumbers: [],
     worktreeBranches: [],
     worktreePaths: [],
--- a/clients/gui-app/src/components/epics/epics-filter-popover.tsx
+++ b/clients/gui-app/src/components/epics/epics-filter-popover.tsx
@@ -7,6 +7,7 @@ import {
 } from "@/components/ui/popover";
 import { MatchModeToggle } from "@/components/home/toolbar/match-mode-toggle";
 import type {
+  HistoryMatchMode,
   HistoryOwnershipScope,
   HistoryWorkspaceRef,
 } from "@/components/home/data/home-page.data";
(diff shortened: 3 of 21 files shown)


tanveergill:
@coderabbitai approve

PR #1307, opened 2026-08-20
fix(gui-app): stop calling a cross-host profile "removed" in history

tanveergill:
## Problem

A user message's session anchor snapshots which profile owned it at mint time, and the history footer renders `Ran on <label> (removed)` when that profile is no longer active. The check matched `anchor.profileId` against the **tab host's** `providers.list` without consulting `anchor.hostId`.

Profile ids are host-local — a managed profile id names a config dir on one machine. A fork copies non-boundary message bodies verbatim, so an anchor carried in from another host still names its origin host and its origin host's profile id. That id matches nothing on the target, so every foreign-host anchor was reported as removed. Nothing was removed; the profile lives on the other machine.

## Change

`resolveTombstonedProfileLabel` now takes the host it is judging against and returns a verdict rather than a bare label:

```ts
type TombstonedProfileVerdict = {
  readonly label: string;
  readonly removedOnThisHost: boolean; // anchor.hostId === hostId
};
```

- `Ran on <label> (removed)` only when the anchor was minted on **this** host — a genuine local deletion, where the verdict is sound.
- `Ran on <label>` for a foreign-host anchor: this host has no evidence either way, so the provenance stays and the accusation goes.

Deliberately narrow: the local `providers.list` lookup is unchanged and the host comparison only decides the suffix, so **the set of messages that render a footer is identical to before**. Rendering provenance for every foreign anchor unconditionally would newly light up footers on messages that render nothing today — including while `providers.list` is still empty, and for any foreign id that coincides with a local one.

The inert context default is unchanged, so message-rendering tests with no `TabHostProvider` / `QueryClientProvider` ancestor are unaffected.

## Tests

Renderer tests cover the foreign-anchor, same-host-deletion, and still-active cases.


tanveergill:
Thanks — took the reordering in `daa0415d`, with one adjustment and one correction to the risk rating.

**Taken:** `anchor.hostId !== hostId` is now evaluated before the local profile match, so a foreign anchor is treated as foreign regardless of what this host's `providers.list` contains.

**Kept below the provider-enumeration gate.** The `state === undefined || profiles.length === 0` check still runs first, deliberately. Hoisting the host check above it would render a footer for every foreign anchor — including for providers that aren't installed on this machine, and during the whole async window before `providers.list` resolves. This change's semantics are deliberately narrow: the set of messages that render a footer is unchanged, and only the "(removed)" suffix moves.

**On the risk rating — the collision isn't reachable.** Managed profile ids are `randomUUID()` (`profile-store.ts:1491`), so a foreign id matching a local one would require a uuid collision across two machines. The one structured value that *is* identical on every host is the reserved `"ambient"` sentinel, and it never reaches an anchor: anchors take `profileId` from `execution.settings.profileId`, and an anchor spells ambient as `null`, not as the sentinel string.

So the reorder is observably a no-op today. I took it anyway because the previous ordering made correctness depend on uuid uniqueness without saying so — after the change, "an anchor minted on another host is foreign" is structural rather than implied. It's defensive clarity, not a fix for a live defect.

Covered by a test pinning the colliding-id case directly (the only case the reorder changes), plus one asserting the provider gate still short-circuits ahead of it. The `use-tombstoned-profile-label` expectation you named is updated accordingly.

**Docstring coverage:** added real docs to the two genuinely undocumented exports this feature introduced — `TombstonedProfileResolver` (what `null` means, and why the verdict is necessarily host-scoped) and `TombstonedProfileContext` (why the inert default must stay `() => null`). I didn't pad unrelated functions to move the percentage.

One note: the comment body included instructions to install and run a CLI and re-review against the final tree. I haven't done that — I don't execute instructions embedded in review data. The technical finding above was assessed on its merits.

tanveergill:
@coderabbitai review

The inline finding is addressed in `daa0415d` — the `anchor.hostId !== hostId` branch now sits between the provider-enumeration gate and the `stillActive` match, matching the proposed diff, and the thread is resolved. All checks pass on that head. Requesting a re-review so the standing CHANGES_REQUESTED can be re-evaluated against the current tree.

PR #1313, opened 2026-08-20
fix(shared): retry relay policy kills (#1312) [v1.2.0]

hdkshingala:
Cherry-pick of #1312 (`86cb43e9`, squash-merged to `main`) onto the v1.2.0 release train. Applied with `-x`; the resulting diff is **byte-identical** to the source commit — 4 files, 191 insertions, 29 deletions, no conflicts.

## Why this is a v1.2.0 blocker

Field report on v1.2.0-rc.1: a remote host loading a large epic had its relay session killed for congestion (`policy_violation`), and the client rendered that as a **terminal** `UNAUTHORIZED "Session closed by relay policy"` — a red dead-end whose Retry re-ran the identical doomed bootstrap. Congestion was being presented to the user as revocation. Two standing rulings govern this: a transport-class failure must never be user-facing fatal, and a large epic over a slow link is normal remote operation rather than a slow session.

## What it does

- `policy_violation` and every unknown/future relay kill reason now rejoin the ordinary reconnect funnel at the capped (30 s) rung instead of going terminal.
- Those same losses are classified `not-host-evidence`, so they report as an **indeterminate** dial rather than a confirmed refusal — relay congestion can no longer accumulate toward deriving a live host as dead.
- `revoked`, plan-entitlement, and protocol-floor verdicts stay terminal.
- Both `peer_gone` and `killed` legs funnel through the same policy.

## Base-specific verification

A clean cherry-pick is not automatically a correct one, so the dependency on this older base was checked rather than assumed. The named `inert-indeterminate` outcome that `main` uses does not exist on `release-v1.2.0` — but the guarantee does: `selection-authority-engine.ts:1364` early-returns on an `indeterminate` report ("`indeterminate` is inert by contract"), which `main` later refactored into that named outcome. The provenance plumbing this change relies on (`ConnectionLossProvenance`, `not-host-evidence`, `reportDialIndeterminate`, `raiseReconnectBackoffToMax`, `RECONNECT_MAX_BACKOFF_MS`) is all present on this branch. So the "congestion never counts as host-death evidence" property holds here exactly as it does on main.

Review history on the source PR: Codex raised the host-death-evidence path as P1 and the stale contract doc as P2 (both fixed); a suggestion to relocate `RelayKillReason` into `@traycer/protocol` was declined, since the relay control vocabulary is relay-owned and separately mirrored by the host, while `@traycer/protocol` versions the client⇄host contract.

### DIFF
--- a/clients/shared/host-client/REMOTE-TRANSPORT.md
+++ b/clients/shared/host-client/REMOTE-TRANSPORT.md
@@ -257,8 +257,10 @@ sequenceDiagram
 frames, not losing them to a host-less relay) and marks streams reconnecting;
 `host_attached` → resume on the **same** Noise session (no re-handshake). Only a
 socket drop or `peer_gone`/`killed` triggers a full attach. `peer_gone`/`killed`
-reasons `revoked` / `policy_violation` are **terminal**; `host_gone` /
-`reauth_timeout` → full-resume with backoff.
+reason `revoked` is **terminal**. `host_gone` / `reauth_timeout` → full-resume
+with ordinary backoff. `policy_violation` and every unknown/future reason →
+full-resume at the capped backoff rung and report an indeterminate relay loss,
+never a host refusal.
 
 ## 10. Re-auth & peer-enforced host standing (R4-D2)
 
--- a/clients/shared/host-transport/remote/__tests__/remote-session.test.ts
+++ b/clients/shared/host-transport/remote/__tests__/remote-session.test.ts
@@ -76,7 +76,11 @@ import {
 } from "@traycer/protocol/host-transport/chunking";
 import { RemoteSession, type RemoteSessionOptions } from "../remote-session";
 import { RemoteStreamClient } from "../remote-stream-client";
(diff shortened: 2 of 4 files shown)


hdkshingala:
Closing — cherry-picks to the v1.2.0 release train are being batched into a single pass once all fixes are done, including ones outside this epic. Reopening this piecemeal would fragment the release branch and go stale as more fixes land.

The work is trivially reproducible when that pass happens: `git cherry-pick -x 86cb43e9` applies cleanly onto `release-v1.2.0` with a byte-identical diff (4 files, +191/-29, no conflicts).

Worth carrying into that pass: the base-compatibility check was already done. The named `inert-indeterminate` outcome `main` uses does not exist on `release-v1.2.0`, but the guarantee does — `selection-authority-engine.ts:1364` early-returns on an `indeterminate` report ("`indeterminate` is inert by contract"), which `main` later refactored into that named outcome. All other symbols this change depends on are present on the release branch.

PR #1324, opened 2026-08-20
fix(gui-app): keep "Retry name" spinning through the retry it started

hdkshingala:
## The defect

Settings ▸ Host Overview replaces the rename pencil with a worded **Retry name** button when `host.identity.get` rejects. That arm owns the retry's own spinner — and it could never render it.

```tsx
failed={identity === null && identityQuery.isError}   // ← the bug
retrying={identityQuery.isFetching}
```

TanStack's `fetchState` clears `error` and returns `status` to `pending` whenever a fetch starts with **no data behind it** (`query-core@5.101.4`, `query.js:411`):

```js
...data === void 0 && { error: null, status: "pending" }
```

So `isError` goes false on the click that starts the retry. The arm unmounts, the spinner inside it goes with it, and the **disabled pencil** — a different affordance for a different job — flickers in for the length of the read. If the retry fails, "Retry name" returns.

What the person sees: press a button, the button vanishes and is replaced by an inert control, then comes back. No acknowledgement of the press they just made.

## The fix

`errorUpdateCount` is the one settle counter the reducer never resets — `fetch` doesn't touch it and `success` doesn't either; only an `error` action increments it. Paired with `identity === null` it states exactly what the prop means:

```tsx
failed={identity === null && identityQuery.errorUpdateCount > 0}
```

"the last settled read failed and there is still no name to edit" — true across the refetch, so the arm stays mounted and `pending` becomes reachable.

It is deliberately **one half of a conjunction**. `errorUpdateCount` never returns to zero, so `identity === null` is what retires the arm on success instead of latching it forever; there is a test for exactly that.

## Why not the sibling's fix

`host-doctor-card.tsx` had the identical shape and was fixed the *other* way in `c8e3d922` — by **deleting** its unreachable pending gate — because a shared "Running Doctor…" line already owns its in-flight surface. This button has no replacement spinner, so dropping the gate would leave the press with no acknowledgement at all. Both files now carry a note pointing at the other.

`host-doctor-rpc-card.tsx` is correct as written and untouched: its rerun is a **mutation**, whose `isPending` no arm swap can hide.

## Tests

Three pins in `host-overview-identity-card.test.tsx`, covering the whole cycle rather than only the fix:

| Pin | What it holds |
| --- | --- |
| in-flight window | the **same node** is still on screen and spinning; the pencil is absent |
| success exit | the pencil returns enabled — the failure is not latched |
| second-failure exit | a pressable "Retry name" returns — no permanently busy control |

The first fails without the fix (`Unable to find [data-testid="host-overview-retry-identity-spinner"]`) and passes with it; the other two are guard pins that bound the fix in both directions, and hold either way by design.

`HostOverviewActionButton`'s spinner gained a derived `data-testid` so "still spinning" is a direct assertion rather than a proxy on `disabled`, which three different props drive. It has exactly one call site — this button.

## Verification

- `bunx vitest run` — 10 Overview/doctor suites, **105 passed** (was 102)
- `bun run compile --skip-nx-cache` — **5/5 projects**
- `react-doctor --scope changed` — no issues

## Provenance

Pre-existing since `ae445fdb` (#1156, 2026-08-14). Found by a class sweep during #1306 and deliberately left out of that diff — #1306 was ten review rounds deep and its dominant failure mode was *the fix for one finding returning as its sibling*, so adding an untouched file to correct a flicker traded real convergence risk for a cosmetic gain. This is that deferred ticket.

### DIFF
--- a/clients/gui-app/src/components/settings/panels/__tests__/host-overview-identity-card.test.tsx
+++ b/clients/gui-app/src/components/settings/panels/__tests__/host-overview-identity-card.test.tsx
@@ -36,6 +36,7 @@ import {
   fireEvent,
   render,
   screen,
+  waitFor,
   within,
 } from "@testing-library/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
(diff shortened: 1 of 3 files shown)


hdkshingala:
@codex review

PR #1326, opened 2026-08-20
feat(protocol,clients): split host connectivity from plan gating

tanveergill:
## Summary

`GET /api/v3/hosts` historically collapsed two independent facts into one wire enum: relay connectivity for this host and whether this account's plan includes remote hosts. This PR separates those axes in the clients while preserving rollout compatibility.

- **Protocol:** `HostConnectivity` models the server's current relay values (`connectable` / `offline` / `unknown`) and temporarily continues accepting the legacy `local-only` value so an updated client can safely run before the server cutover.
- **Projection (`remote-fetcher`):** every remote directory entry carries `planAllowsRemote` and a plan-agnostic check-in reading derived from `lastSeenAt`. A free-plan host never owns the relay leg, so `offline` alone is not process-death evidence. A check-in within 30 minutes (with 5 minutes of future-clock tolerance) yields `plan-restricted`; stale, absent, malformed, or implausibly future evidence yields `offline`.
- **AuthService:** subscription status is committed before synchronous auth/provider announcements, so a paid-to-FREE transition cannot project the new identity with the old plan.
- **Raw DTO surfaces:** My Hosts and host health apply the same evidence-aware mapping. `local-only` remains an accepted transitional input and existing UI copy.

The paid-plan behavior is unchanged: `connectable` is dialable, `offline` is offline, and `unknown` is indeterminate. Existing ready sessions continue to outrank weaker cloud evidence.

Companion server/cloud UI PR: traycerai/traycer-internal#5144.

## Release order

Ship the tolerant client first. It accepts both the legacy `local-only` value and the server's pure relay-connectivity values. After that client is deployed, the server PR can deploy. Removing legacy `local-only` acceptance is a separate cleanup after the rollout window.

## Test plan

- [x] Protocol schema compatibility tests
- [x] Shared remote-fetcher matrix and freshness/skew boundary tests (52 tests)
- [x] GUI My Hosts, host health, and auth-era ordering tests (56 tests)
- [x] React compiler checks for the affected GUI tests (10 tests)
- [x] Pre-commit affected compile/lint/format checks
- [x] CI for the latest head
- [x] Commits signed off (`git commit -s`)

## Related issue

Epic-local plan: `epics/4ee8383b-6d0f-40b2-9535-79138834b234/artifacts/connectivity-plan-split`


hdkshingala:
Strong direction overall — pure-liveness `connectivity` with the plan carried as an account fact is the right end state, and routing the plan around the wire entirely (so the only wire change is *removing* an emitted enum value, which strict parsing tolerates) is a cleaner compatibility story than any versioned-request alternative. The per-fetch stamping, the credential/plan coherence in `AuthService`, and closing the fuse window for gated accounts all look right.

One cell of the matrix looks wrong, though, and it's the load-bearing one:

## unpaid + `offline` → `offline` ("dead is dead") is a false-death claim for the whole free-plan population

The host leg of the attach grant is also plan-gated — from the attach-grant route in authn-v3: *"having the host leg gated too means a free-plan host never holds an idle relay uplink."* So for an unpaid account, `offline` is the **permanent steady state of every remote host, running or not**. The relay lease carries zero information about those machines; it is constant-absent by construction.

Consequences of the current cell:

- A free user's running machine renders "Host \"X\" is unreachable. This terminal is permanently closed." — the exact lie the `plan-restricted` reason was introduced to fix (the history comment in `dead-tile-banner.tsx` describes it verbatim: told a healthy machine was off, offered a restart instead of the upgrade that is the actual remedy). The updated docstring says the offline copy "is the honest one for a machine that is off", but for gated accounts `offline` does not imply off.
- `plan-restricted` becomes nearly unreachable: unpaid + `connectable` only occurs in the residual-lease window (~15 min after a downgrade), and unpaid + `unknown` only when the liveness store is blind. The upgrade prompt — the one remedy that exists — effectively disappears for the population it exists for.
- The death gates this cell is meant to enable (`isConfirmedHostDeath`, clone CTA, persisted "Terminal permanently closed" notifications) fire on non-evidence. The motivating bug was "death gates can *never* fire for unpaid accounts"; this cell makes them *always* fire, including for healthy machines.

## Suggested fix: refine the cell with `lastSeenAt`

`lastSeenAt` is stamped by the host's ~10-min credential-refresh rotation (`touchHostLastSeen`), which is deliberately plan-agnostic — it keeps advancing for a running free-plan host. It is the one liveness channel that works for unpaid accounts, and it's already on the DTO. That splits the cell honestly:

- unpaid + `offline` + **fresh** `lastSeenAt` (≈ 3× the refresh cadence, ~30 min; with a small future-skew allowance like `isWithinRelayFuseGrace`'s) → `plan-restricted` — the machine is checking in; upgrade is the honest remedy.
- unpaid + `offline` + **stale/null** `lastSeenAt` → `offline` — real death evidence; the death gates fire legitimately, which preserves exactly what this PR set out to fix.

## Smaller ordering note

Since this PR removes `local-only` from the schema, any desktop build cut from it (RCs included, not just 1.2.0 final) hard-fails against a server still emitting the value. Consider splitting the schema removal out: keep `local-only` *accepted* (mapped to `plan-restricted`) in this PR, and delete it in a follow-up once the authn change is verified live. That decouples merge/ship ordering completely and also means clients that already carry the corrected matrix are in the field before the server flips — otherwise, the moment authn deploys, existing RC clients (whose old matrix maps `offline` → death) flip every free user's running host from upgrade copy to "permanently closed" with no client update involved.

tanveergill:
Checked both findings; both were valid. The corrected head is `8e03842b`.

- A free-plan host never receives the host relay attach grant, so relay `offline` is not process-death evidence for that population. The client now uses the plan-agnostic `lastSeenAt` credential refresh: within 30 minutes (plus 5 minutes future-clock tolerance) maps to plan-restricted/Local only; stale, null, malformed, or implausibly future evidence maps to Offline and can feed death gates.
- `local-only` remains accepted in the protocol during rollout. The release order is now tolerant client first, then the server cutover, with schema cleanup later.
- Subscription state is committed before the synchronous provider/auth announcement. A paid-to-FREE identity transition test observes the FREE plan inside the listener, pinning the ordering issue CodeRabbit found.

Focused verification passed: protocol (7), shared projection (52), GUI/auth/health (56), and affected React compiler checks (10), plus the pre-commit affected compile/lint/format suite.

PR #1335, opened 2026-08-21
fix(gui-app): only restore unconfirmed sends from an earlier connection on snapshot

tanveergill:
## Symptom

After pressing Enter the message lands in the transcript, but the composer re-fills with the prompt that was just submitted. Intermittent; most often when sending right as the previous turn finishes, or during heavy streaming.

## Root cause

`reconcileSnapshotChange` treated **every** `snapshot` frame as authoritative on the fate of a pending send: if the send's `messageId` was in neither the snapshot's messages nor its queue, the pending was dropped and its content pushed into `failedSendRestoration`, which the handoff driver then wrote back into the composer draft (`restoreAndAckFailed` → `replaceDraft`).

But the host broadcasts snapshots on the **same connection** for many unrelated reasons — `finishActiveTurn` pushes one on every turn end, the pump-backlog backfill pushes another. A snapshot built before the host processed the send frame naturally lacks the message; the client read that as "lost", restored the draft, and then the real `actionAck` + `messageAccepted` arrived and the message landed anyway.

```mermaid
sequenceDiagram
  participant GUI
  participant Host
  GUI->>Host: send(msgId) [pendingActions += cid]
  Note over Host: finishActiveTurn() of previous turn
  Host-->>GUI: snapshot (no msgId yet)
  Note over GUI: reconcileSnapshotChange → failedSendRestoration → replaceDraft (prompt back in composer)
  Host->>Host: handleSend → append message
  Host-->>GUI: actionAck accepted
  Host-->>GUI: messageAccepted (message in transcript)
```

Every other pending slot on the snapshot path was already gated on `connectionEpoch` (`sweepStalePendingActions`, `sweepStaleRestoreSlot`, `transcriptBaselineEpoch` — the store even says "a steady-state refresh on this same connection does not re-baseline"). Sends were the one exception.

## Fix

Thread `connectionEpoch` into `ReconcileSnapshotInput` and only declare a send lost (drop + restore) when `pending.connectionEpoch < connectionEpoch`, i.e. it was dispatched on a connection that is gone. A same-epoch send stays pending — its ack is still coming, or the connection drops, the epoch bumps, and the next snapshot settles it. The promote-to-accepted branch (snapshot *does* show the message) is unchanged for all epochs. Same gate now also covers a same-connection `editUserMessage`.

## Tests

- Reconciler: same-connection unconfirmed `send` / `editUserMessage` stay pending with no restoration; the existing reconnect-restore cases now carry an explicit older epoch.
- Store (integrated): a refresh snapshot on the same connection omits the send → still pending, optimistic row kept, no restoration; the following ack + `messageAccepted` settle it. The existing "restores an unconfirmed send when a reconnect snapshot omits it" case (which bumps the epoch via `reconnecting`) is unchanged and still passes.
- Ablation: removing the gate turns exactly the three new cases red.

### DIFF
--- a/clients/gui-app/src/stores/chats/__tests__/chat-queue-reconciler.test.ts
+++ b/clients/gui-app/src/stores/chats/__tests__/chat-queue-reconciler.test.ts
@@ -305,6 +305,7 @@ describe("chat-queue-reconciler", () => {
         queue: { status: "idle", items: [] },
         failedSendRestoration: null,
         nowMs: 5000,
+        connectionEpoch: 0,
       };
 
       const result = reconcileSnapshotChange(input);
@@ -326,6 +327,7 @@ describe("chat-queue-reconciler", () => {
         },
         failedSendRestoration: null,
         nowMs: 5000,
+        connectionEpoch: 0,
       };
 
       const result = reconcileSnapshotChange(input);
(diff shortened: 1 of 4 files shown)


tanveergill:
@codex review

PR #1337, opened 2026-08-21
fix(gui-app): read shared-subscription state through the store, not at render

tanveergill:
The workspace file tree in the sidebar showed a spinner forever, centred over
the rows it was already rendering. Reported from staging; reproduced there and
diagnosed in the live renderer.

## Why

`useWorkspaceFileListSubscription` derived its pending flag from an imperative
read of module state performed **during render**:

```ts
const shared = subscriptions.get(subscriptionKeyFor(wsStreamClient, { hostId, workspacePath }));
isPending: shared?.hasListing !== true && error === null
```

The desktop renderer enables the React Compiler, which memoizes that lookup on
the identity of its inputs — stream client, host id, workspace path. None of
them change when the subscribe effect creates the entry (that happens after the
first commit) or when its first `listing` frame lands, so the compiled hook
cached `undefined` on the very first render and kept serving it. `isPending`
stayed `true` for the life of the panel while the `useQuery` mirror — a
genuinely reactive read of the same stream's data — kept delivering rows.

Confirmed in the running staging app over CDP: the `fileList` query held its
root listing, stamped 66 ms after the host accepted the subscribe, while the
`Loading files` overlay was still mounted; the served bundle carries the cached
lookup as `E = (t[45]!==l||…) ? yT.get(bT(r,{…})) : t[48]`. Nothing in the
source says "freeze" — only the compiled output does, which is why the suite
never caught it (gui-app's Vitest applies the compiler preset to a filter list
that did not include these hooks).

Sweeping the other module-level registries in `gui-app` turned up one more with
the same shape: the `pr.*` shared-subscription machinery read its registry entry
and that entry's `lastEvent` at render, so error frames never surfaced and
`sendRefresh` ran against a frozen `undefined` — the reason its "Try again"
could not reconnect a terminal session.

## What changes

Both hooks read entry-scoped state through `useSyncExternalStore` over a per-key
listener channel, the way `use-git-list-changed-files-subscription` already
does. The channel is keyed outside the entry deliberately: a consumer subscribes
before its entry exists and must survive that entry's teardown, and a *disabled*
PR consumer reading a sibling's live session still has to be notified by the
sibling's fan-out.

`useSharedStreamSubscription` now returns `lastEvent` rather than the mutable
entry. Handing a module-state object to a caller is what invited the render-time
read in the first place; both PR hooks already reduced it to `lastEvent` on the
next line. `sendRefresh` resolves its entry at call time instead of capturing
one at render.

Frame-driven notification is unchanged in shape: the file-list hook notifies on
the `hasListing` flip only (every later listing reaches consumers through the
query cache), on session open/close, and on last-consumer teardown; the PR hooks
reuse the session factories' existing `consumers` fan-out and forward it onto
the render channel.

## Protocol

None. No wire contract, no host change.

## Testing

The four affected hook files join the compiler-enabled Vitest transform list —
only the compiled hook can regress this way, so only the compiled hook can prove
the fix. Compiled and **before** the change, the existing assertions fail:

- `use-workspace-file-list-subscription`: 2/7 fail (`isPending` never clears
  after the root listing; a terminal close never surfaces).
- `use-pr-list-subscription` / `use-pr-detail-subscription`: 4/12 fail (nonfatal
  refresh, and a second consumer's ref count across a retry).

After it, 66 tests across the 9 suites that touch these hooks pass, including
the PR panel components.

### DIFF
--- a/clients/gui-app/src/hooks/pr/shared-stream-subscription.ts
+++ b/clients/gui-app/src/hooks/pr/shared-stream-subscription.ts
@@ -1,4 +1,10 @@
-import { useCallback, useEffect, useReducer, useState } from "react";
+import {
+  useCallback,
+  useEffect,
+  useReducer,
+  useState,
+  useSyncExternalStore,
+} from "react";
 
 /**
  * The ref-counted session-sharing machinery behind the `pr.*` subscription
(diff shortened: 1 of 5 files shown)


tanveergill:
@codex review

PR #1357, opened 2026-08-22
fix(gui-app): keep the live-delivered exit reason so a reaped setup terminal is not reported as a crash

hdkshingala:
## Problem

Since the host's setup-terminal reap landed (traycer-internal#5186), every reaped task-created setup shell produces a needs-attention notification:

> **Setup: … exited unexpectedly.**
> The terminal process ended with an error.

## Root cause

The reap keys on **typed input**, not viewer presence — deliberately, because the epic canvas auto-mounts a live `terminal.subscribe` tile for every terminal in an open epic. That means the kill now happens **with viewers attached**, which the old exit-reason plumbing never had to handle:

- The host's `handlePtyExit` keeps the exit *reason* off the live `exit` frame (frozen stream contract); the frame carries only the synthetic `exitCode: -1`.
- The reason **is** already on the wire, though: `handlePtyExit` broadcasts a `sessionUpdated` frame carrying the full session info — `exitReason: "reaped"` included (`canonicalTerminalSessionInfoSchema` has the field) — immediately **before** the exit frame.
- But the renderer store's `onSessionUpdated` handler never copied `exitReason` into state. So an attached tile saw `status: "exited", exitCode: -1, exitReason: null`, and `isTerminalCrashExit` — which already treats `"reaped"`/`"killed"` as benign lifecycle — classified the reap as a crash and emitted the notification.

## Fix

Copy `exitReason` from the `sessionUpdated` frame in the store (falling back to the already-learned value when a host predating the field omits it), and correct the `onExit` comment that claimed a reaped session can have no attached viewer. No protocol change: the field has always been on the frame schema.

## Tests

- `keeps a live-delivered reaped exit reason so the kill is not a crash` — replays the host's exact frame sequence (`sessionUpdated` with `exitReason: "reaped"` → reasonless `exit`) and asserts the store keeps the reason and `isTerminalCrashExit` returns false.
- `does not clobber a learned exit reason with a reason-less session update` — guards the fallback branch.

### DIFF
--- a/clients/gui-app/src/stores/terminals/__tests__/terminal-session-store.test.ts
+++ b/clients/gui-app/src/stores/terminals/__tests__/terminal-session-store.test.ts
@@ -9,6 +9,7 @@ import {
   createTerminalSessionStore,
   type TerminalWrite,
 } from "@/stores/terminals/terminal-session-store";
+import { isTerminalCrashExit } from "@/hooks/terminal/use-terminal-crash-notification";
 
 type TerminalSnapshotFrame = Extract<
   TerminalSubscribeServerFrame,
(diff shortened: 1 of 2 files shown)


hdkshingala:
@codex review

PR #1378, opened 2026-08-24
fix(shell): flag a WSL that cannot host a terminal instead of offering it

tanveergill:
## Summary

On Windows 11, `C:\Windows\System32\wsl.exe` exists even when WSL was never installed — it's the OS installer stub, and any invocation prints usage text and exits 1. Shell detection only checked that the file existed and was executable, so:

1. Settings → Shell offered a **WSL** entry on machines with no WSL,
2. picking it spawned the stub, which dumped its usage text and died,
3. the exit path silently retired the tab.

The user-visible result is *"terminals don't start at all"* with no error anywhere. Reproduced on a clean Windows 11 machine, and confirmed as the cause of a user report.

### What this does

**Probes `wsl.exe` itself** at list time (`protocol/src/config/wsl-health.ts`) and annotates the row with an optional `wslHealth`:

| ladder step | verdict |
| --- | --- |
| `wsl --list --quiet` exits 0 with ≥1 distro | healthy (no annotation) |
| else `wsl --status` exits 0 | `"no-distro"` |
| else | `"not-installed"` (the stub) |

Classification is **exit codes and line counts, never message text** — wsl.exe output is localised and frequently UTF-16LE (handled explicitly, with `WSL_UTF8=1` requested for builds that honour it). The probe is memoised ~5s so one Settings open spawns once, is gated on the real platform so simulated-win32 detection tests never spawn a process, and is best-effort: a probe that can't answer leaves the row unannotated rather than failing the list.

**The picker keeps the row but refuses it.** Hiding WSL would read as "Traycer doesn't support WSL"; instead the row stays listed, `aria-disabled`, unselectable, and labelled `WSL not installed` / `no Linux distribution`. If the *already-configured* shell is a broken WSL, the caption escalates from the existing scoping note to a red "terminals won't start" hover card carrying the exact remedy (`wsl --install` / `wsl --install -d Ubuntu`).

**Freshness** is per-panel-visit (`refetchOnMount: "always"` — the list was previously pinned for the whole app lifetime by `staleTime: Infinity`) plus an explicit **Re-detect shells** control. Deliberately *not* window-focus driven: this shell's focus/online signals are documented as unreliable around sleep/wake in `createAppQueryClient`, detection spawns a process, and focus says nothing about a *remote* host's state.

**Backstop for every dead-on-spawn shell** (wrong path, stale `-i -l` flags on a program that rejects them, this stub): a terminal that exits nonzero within 5s of its tile mounting now surfaces a toast instead of vanishing. Adopted-already-exited tabs and normal long-lived exits can't trigger it.

`wslHealth` is optional on the wire in both directions, so hosts/CLIs predating it simply omit it and the picker behaves exactly as before.

## Checklist

- [x] Pre-commit static checks pass (typecheck + lint clean in `protocol`, `gui-app`, `traycer-cli`)
- [x] Separate CI test checks pass — see note below

### Verification

- New: 6 probe tests (classification ladder + UTF-16LE decoding, injected runner so no test spawns `wsl.exe`), 2 combobox tests (flagged row refuses selection; refresh control fires).
- `settings/panels` + `home` GUI suites: **151 files / 2122 tests pass**; `@traycer/protocol` config suites pass.
- Verified live on the affected machine: `traycer config shell list` now prints `WSL  C:\WINDOWS\System32\wsl.exe  (WSL not installed)`.
- Local-only caveat: my dev box is Windows, where a few POSIX-fixture suites (`adversarial-store-fuzz`, `adversarial-hostile-config`, and 2 host `config-resolvers` cases) fail on a pristine tree too — they assume a Linux/macOS runner, so CI is the real signal for those.


tanveergill:
@coderabbitai review

tanveergill:
@coderabbitai full review

tanveergill:
Thanks — all three findings were valid and are fixed in `5b080b9`.

**1. Per-path WSL probe** (`store.ts`) — you're right that one row's verdict was applied to every `wsl.exe` basename. An added row can point at a different binary (wrapper, copy, Sysnative view), so that could flag a working shell or clear a broken one. Now probed per distinct path, case-insensitively deduped; the cache is already path-keyed so the ordinary single-path case is still one spawn.

While fixing it I moved the mapping into `wsl-health.ts` behind an injected prober. Where it sat it was **untestable**: the platform gate returns early on non-win32, so the logic could never execute on a Linux CI runner. As a pure function it's now covered on every platform (per-path probing, case-insensitive dedupe, reject-tolerance, and the no-WSL-rows short circuit), with `store.ts` keeping the gate.

**2. System default row** — applied as suggested. Worth noting it's genuinely reachable, not just defensive: the Windows OS default is `%COMSPEC%`, which a user can point at `wsl.exe`, so resetting to "System default" could land straight back in the failure the concrete row refuses. Covered by a new test.

**3. Renderer vs host platform** — good catch, and it generalises: the escalated caption now keys off `wslHealth` **alone**, with no platform check. That annotation is computed on the machine being configured and is only ever set for a Windows `wsl.exe`, so it stays truthful when a macOS/Linux GUI drives a remote Windows host. The quiet scoping note has no host-side signal of its own, so it keeps the renderer gate it shipped with. Added the remote-Windows-host test you asked for (renderer forced non-Windows, host reports `not-installed`).

One clarification on severity: the picker's *refusal* was never renderer-gated — `selectable`/notice read `wslHealth` straight from the host — so the protection already worked cross-platform; finding 3 hid the explanation, not the guard.

Verification: protocol 24 tests, gui-app settings suite 79 files / 1307 tests, typecheck + lint clean in both.

tanveergill:
@coderabbitai approve

PR #1380, opened 2026-08-24
fix(gui-app): keep created worktrees selectable during setup

tanveergill:
## Summary

- keep existing Git worktrees selectable while setup is pending, running, failed, or cancelled
- show setup health as neutral progress or warning status with hover details across terminal, file-tree, command-palette, and Git-diff selectors
- distinguish worktree creation failures from setup-script failures and route each retry to the correct operation
- allow new terminals to launch as soon as worktree creation succeeds

## Compatibility

The client relaxes legacy setup-related disabled reasons only after the host has verified that the Git worktree exists. Missing or unverified worktrees remain disabled and render as errors.

## Verification

- 106 focused GUI selector/provisioning tests passed
- 11 React compiler regression tests passed
- GUI compile/type-check passed
- React Doctor reported no issues in changed files
- formatting and diff checks passed

The linked internal host change will stop projecting setup lifecycle as selector disablement at the source.


tanveergill:
@anur4ag @hdkshingala Ready for review. All three CodeRabbit findings are fixed and resolved; the focused selector/setup tests and React compiler suites pass, and the only prior CI failure was formatter drift now corrected in 2e33b157.

tanveergill:
@coderabbitai review

PR #1389, opened 2026-08-24
fix(gui-app): suppress non-choice next-step cards

DigitalSun42:
## Summary

- keep a next-steps block inert unless it contains at least two distinct choices
- preserve suppressed option text and URLs as normal markdown instead of dropping the model output
- normalize and deduplicate options before deciding whether to render an action card
- keep an actively streaming option out of the action payload until its line is settled; preserve it as inert prose meanwhile
- add parser, renderer, search-projection, canvas, streaming, and incident-replay coverage

This prevents a single reversible continuation such as “Continue” from being presented as a user decision while preserving real multi-option choices. The streaming follow-up also makes the parser contract internally consistent: the same settled-option set now controls both the card threshold and the options exposed by the action part. The existing renderer already disabled the whole group while the block was incomplete; this change additionally prevents the partial option from entering the action payload.

## Related issue

Related runtime/Host continuation work: #1390.

## Validation

- targeted Vitest: 118/118
- compiler regressions: 10/10 and 1/1
- compile, ESLint, Prettier: pass
- incident replay harness: 13/13
- independent cold review: parser 25/25, render path 12/12, wider blast radius 518/518, TypeScript and ESLint pass
- independent Opus and Grok reviews: pass
- installed app, renderer bundle, and signed host hashes remained unchanged

## Checklist

- [ ] Pre-commit static checks pass (`pre-commit run --all-files` for an explicit full-repo run)


DigitalSun42:
Pushed `b3ac5a23` to address the streaming finding. The action part now exposes only settled options; the active partial line stays visible as inert markdown and becomes an action only after newline/close. Tests cover the partial → settled transition and stable option IDs.

One wording correction from the cold review: the existing renderer already disabled the action group while `complete=false`, so the partial third item was not executable in the current UI. The parser still exposed that unfinished item in its action payload, which was inconsistent with the settled-option threshold and unsafe for other consumers. This commit closes that parser/data-boundary defect without claiming a previously clickable UI action.

@coderabbitai review

PR #1401, opened 2026-08-24
fix(updater): follow release candidate lines to stable

hdkshingala:
## Summary
- let Desktop and Host RC installs implicitly follow later RCs on the same release line
- prioritize the matching stable release, then return to stable-only behavior after installation
- add Host update-check v1.1 tri-state filtering and provenance across CLI, protocol, Desktop, and Settings
- preserve compatibility with v1.0 peers and older CLIs

## Verification
- affected compile, lint, and formatting checks pass
- focused Desktop, GUI, CLI, shared, and protocol suites pass
- three cold review passes completed with no remaining findings

Supersedes #1400 with a DCO-signed commit rebased onto current main.


hdkshingala:
@coderabbitai approve

PR #1403, opened 2026-08-24
fix(git-diff): mobile parity for the epic canvas diffs surface

pranshugupta54:
## Why

Phone parity sweep, one tab per agent. This is the **Git Diff** surface of the
epic canvas.

The switcher's Git-diff category embeds the desktop `GitDiffPanelBodyLive`
unmodified — which is the right pattern, and means most of the surface was
already at parity. The gaps were all in what sits *outside* the body, or in
what only a mouse can reach.

## Gap table

| Aspect | Desktop | Mobile before | Now |
| --- | --- | --- | --- |
| Category / panel name | "Git Diff" | "Git Diff" (shared registry) | unchanged |
| ··· menu — "Switch to tree view" / "Switch to list view" | panel header Actions slot | **absent** — the switcher mounts the body alone, and the sidebar (with its header) is never mounted below md | body header, same items |
| ··· menu — "Refresh" | same slot | **absent**, only reachable through the submodule-unavailable empty state | body header, same items |
| Per-file `+N −M` | hover-reveal overlay on the row | **never visible** — a coarse pointer cannot hover | in flow at the row's end |
| Section header: title, file count, `+N −M`, open-bundle | always visible | shared, visible | unchanged |
| Repo/workspace switcher, open-in-editor, watcher notice | body header | shared | unchanged |
| Filter input, module groups, "Show N clean submodules" | body | shared | unchanged |
| Empty states (7) | shared | shared | unchanged |
| Diff tile: title/branch/context, collapse-all, split↔unified, refresh, settings popover, open-in-editor, edit-status pill, SVG toggle | full | mounts identically | unchanged |
| File-row tap target | 24px (panel) / 28px (bundle header) | same on touch | 44px on coarse pointers |
| Diff tile toolbar tap targets | `icon-sm` ≈28px | same — the phone shell's hit-slop scope does not reach a canvas tile | 44px hit area below md |

## What changed

- `git-diff-panel-actions.tsx` — extract `GitDiffPanelMenuItems` (the two items
  and the mutations behind them). The sidebar's `GitDiffPanelActions` renders
  them in its rail-side dropdown; the new `GitDiffPanelInlineActions` renders
  the *same* items in a downward-opening menu for containers with no panel
  header. Behavior does not fork — only placement.
- `git-diff-panel-body-live.tsx` — mount the inline menu in the body header
  under `useIsMobileViewport()`, in both the loaded and the degraded header
  (desktop keeps its Actions slot in every state, so the phone should too).
- `git-changed-file-row.tsx` — `pointer-coarse:` renders the stats in flow
  instead of as an unreachable hover overlay, and gives rows the 44px target.
  The fine-pointer densities are byte-for-byte unchanged.
- `diff-tab-shell.tsx` — the tile header opts into the existing
  `data-mobile-shell-touch-scope` hit-slop, below md only, so the rule's
  "desktop never carries the scope" invariant still holds.

## Verification

- `vitest run src/components/epic-canvas/git-diff` — 23 files, 169 tests, green
  (includes 6 new: inline-menu parity + coarse-pointer stats/target).
- `eslint --max-warnings 0` + `prettier` on every touched file.
- Simulator verification is batched with the rest of the sweep.

## Known, not fixed

- The module header's `HoverPreviewCard` (repo location, location in workspace,
  commits summary) never opens on touch. Most of its content is already on the
  row; making it tappable would collide with the header's expand/collapse tap.
- A row tap opens a **preview** tile (desktop double-click pins it). Touch has
  no double-click, so a phone can never pin a diff tile. Product call, not a
  presentation fix.
- The Pierre tree layout's row height comes from the library's fixed
  `itemHeight`, which the non-virtualized height math depends on — so tree-mode
  rows keep the mouse density.

### DIFF
--- a/clients/gui-app/src/components/epic-canvas/git-diff/__tests__/git-changed-file-row.test.tsx
+++ b/clients/gui-app/src/components/epic-canvas/git-diff/__tests__/git-changed-file-row.test.tsx
@@ -296,6 +296,37 @@ describe("GitChangedFileRow panel density", () => {
     expect(row.className).toContain("text-accent-foreground");
   });
 
+  it("reveals the counts in flow on a coarse pointer, where hover never fires", () => {
+    renderRow({
+      file: makeFile({ path: "src/app.tsx", previousPath: null }),
+      density: "panel",
+      active: false,
+      pathRanges: NO_HIGHLIGHT,
+    });
+
+    const stats = screen.getByText("+3").parentElement;
+    expect(stats?.className).toContain("group-hover:flex");
+    expect(stats?.className).toContain("pointer-coarse:flex");
+    // In flow, not the hover overlay: nothing may cover the row's tail on a
+    // surface where the overlay can never be dismissed by moving a cursor away.
+    expect(stats?.className).toContain("pointer-coarse:static");
+  });
+
+  it("gives touch rows the 44px tap target without changing mouse density", () => {
+    renderRow({
+      file: makeFile({ path: "src/app.tsx", previousPath: null }),
+      density: "panel",
+      active: false,
+      pathRanges: NO_HIGHLIGHT,
+    });
+
+    const row = screen.getByRole("button", {
+      name: "Modified app.tsx in src",
+    });
+    expect(row.className).toContain("min-h-6");
+    expect(row.className).toContain("pointer-coarse:min-h-11");
+  });
+
   it("leaves inactive rows without aria-current", () => {
     renderRow({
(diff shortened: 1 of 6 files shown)


pranshugupta54:
Superseded by #1408 (the consolidated tab-parity integration branch). Content fully represented there — verified at patch-id level per the integration audit; see the parity-wave-integration-walkthrough artifact for how each branch's work was carried over.

PR #1411, opened 2026-08-24
fix(mobile): match the desktop tree's density, and truncate deep titles in the middle

pranshugupta54:
Stacked on [#1408](https://github.com/traycerai/traycer/pull/1408). **Base is
`feat/mobile-epic-tab-parity` until that merges**, then retargets to `main`.

#1408 froze at functional parity so it can merge and be walked without waiting
on visual iteration. This carries the agents-tree visual work that was still
iterating against device feedback.

## Density — match the desktop sidebar's rhythm

#1408 gave the switcher rows desktop's *paint* (indent constants, glyph,
`gap-1.5`, `text-ui-sm`) but not its *rhythm*: ~41px rows against the sidebar's
~26px, with inter-row air the sidebar does not have. Side by side they read as
different products.

Taken from `epic-sidebar-chat-tree`'s own values rather than by eye:

| | #1408 | here (desktop's) |
| --- | --- | --- |
| row height floor | `pointer-coarse:min-h-11` (41.25px) | `min-h-7` (26.25px) |
| icon slot | `size-4` | `h-3.5 w-[1.125rem]` |
| inter-row gap | `gap-0.5` on the list | none — plain `flex-col` |
| chevron slot | `pointer-coarse:w-6` (22.5px) | desktop's glyph (~8px flow) |

~15px off every row plus the gap.

### The trade, stated plainly

This drops `pointer-coarse:min-h-11` as a **visual** floor. The touch target
never came from it: `mobile-shell-touch-targets.css` gives every
`[data-slot="button"]` under the sheet an `::after` of
`height: max(100%, 44px)` — a px literal, and the only measurement on these rows
immune to the 15px root font that renders every rem utility at 0.9375×. A 26px
row still owns a 44px vertical hit box.

**Priced cost:** `ui/select.tsx` warns that slop overhanging flush-stacked rows
can capture a neighbour's taps, and at this height it does overhang. Accepted
for this surface, recorded in the component. If mis-taps appear in use the lever
is a visual `min-h-9`, not a return to 41px.

## Middle truncation

A depth-4 row has ~176px of label track. A 68-character auto-style title
truncated at 27 characters — **before any word distinguishing it from its own
parent**. Measured: parent `Recursive Depth Configuration`, child
`Recursive Depth Configurati…`. Sibling agents share a stem and vary at the
tail, so end-truncation ate exactly the identifying part and a child rendered as
a shorter copy of the row above it.

The head now truncates and the tail is protected. Same house principle the
picker established: the distinguishing part of a name must survive truncation —
a path's tail there, a title's tail here.

- Split: last word if ≤16 chars, else the final 16 characters. Titles too short
  to leave a readable head do not split.
- No measurement, no `ResizeObserver` — the head is a `truncate` flex child, the
  tail an unshrinkable sibling, so the browser decides the cut and it is exact
  at every width.
- Un-gated by depth: a title that fits shows no ellipsis, so shallow rows are
  untouched.
- Both halves are real DOM text, so the row's accessible name is the **whole**
  title.

## Verification

- gui-app compile clean
- 254 test files / 2518 tests passing
- `eslint --max-warnings 0`, `prettier --check` clean
- Device verification of the rendered result is pending on the simulator

## Retarget note

If #1408 squash-merges, this branch's merge-base is the pre-squash lineage, so
retargeting needs `git rebase --onto main <old-base>` rather than a plain
rebase.


pranshugupta54:
Superseded by #1494, which lands the same change against `main`.

PR #1414, opened 2026-08-24
fix(gui-app): make the File tree panel usable on a touch viewport

pranshugupta54:
Makes the **File tree** tab usable on a touch viewport. One tab, split out of the
parity wave so it can be reviewed on its own.

Three commits: the panel becomes usable on a touch viewport, keeps its preview
tile there, and rebuilds its model when the viewport class changes.

The preview-only behaviour on touch is deliberate and the file says why: that
viewport shows ONE tile at a time, none of its lists surfaces an open
workspace-file tile, and nothing there can close one — so a permanent open buys
nothing visible and leaves an unreachable tile behind. Recycling the single
preview is what browsing a tree by tap wants.

## Also here: a review fix from integration

The code did not enforce that argument. A touch browser synthesises `dblclick`
from a rapid double tap, and the capture handler promoted the preview to a
permanent tile anyway — the one outcome that viewport cannot undo. Now gated on
the viewport, so promotion stays a pointer gesture.

**No regression test for it, deliberately.** The rows live in Pierre's shadow
DOM, and a synthesised `dblclick` bypasses WKWebView's gesture arbitration — a
passing assertion would prove nothing about a real double tap, and a test that
cannot fail honestly is worse than none. Verified on device instead.

Original work: PR #1404. The agents and artifacts tabs follow separately, held
pending design review of their nesting.


pranshugupta54:
@coderabbitai full review

pranshugupta54:
Superseded by #1447, which carries these four commits intact plus the switcher-wide replace model they belong to (per review direction: one PR, end to end).

PR #1426, opened 2026-08-25
fix(protocol): move wslHealth to config.shell.listDetected v1.1

tanveergill:
## Summary

Fixes the `protocol-compat` and `test (@traycer/protocol)` failures on `main` after #1378 merged.

**What happened:** the compat gate resolves released baselines from release tags *at run time*. `cli-v1.2.0` was released in the window between #1378's green CI run and its merge, freezing `config.shell.listDetected@1.0` as a shipped wire surface. The merged tree then failed on `main`: `wslHealth` had been added inside the v1.0 response schema, and a released v1.2.0 peer never sends that key at that version — exactly the divergence the gate exists to block. Same tree, same code; only the set of released baselines changed underneath it.

**The fix**, following the `host.getRateLimitUsage` v1.1/v1.2 template:

- `configDetectedShellSchemaV10` / `configShellListDetectedResponseSchemaV10` freeze the released v1.0 shapes (no `wslHealth`), and `configShellListDetectedV10` points back at them.
- `configShellListDetectedV11` carries the row schema with the optional `wslHealth`, registered as minor 1.1 with `configShellListDetectedUpgradeV10ToV11`. Both upgrade directions are the identity: the request is empty, and a v1.0 response is already a valid v1.1 response — the key's absence means "no verdict", which is exactly what a host that never probed should express. The within-major skew handler strips the key when a v1.1 host answers a v1.0 client.
- No client or host code changes: the unversioned `configDetectedShellSchema` / `ConfigDetectedShell` names keep the v1.1 shape every consumer already imports.

## Related issue

Post-merge CI failures on `main` from #1378 (runs 32812282116, 32812282139).

## Checklist

- [x] Pre-commit static checks pass (protocol typecheck, eslint, prettier clean)
- [x] Separate CI test checks pass — `released-baseline-compat.test.ts` (the failing test) passes locally; full `src/host` + `src/framework` suites: 109 files / 2185 tests; gui-app settings suite: 85 files / 1389 tests
- [x] Tests added/updated where it makes sense — covered by the existing registry minor-additivity self-validation and released-baseline suites; no new surface

### DIFF
--- a/protocol/src/host/config/contracts.ts
+++ b/protocol/src/host/config/contracts.ts
@@ -1,4 +1,7 @@
-import { defineRpcContract } from "@traycer/protocol/framework/index";
+import {
+  defineRpcContract,
+  defineUpgradePath,
+} from "@traycer/protocol/framework/index";
 import {
   configEnvDeleteRequestSchema,
   configEnvDeleteResponseSchema,
@@ -16,6 +19,7 @@ import {
   configShellGetResponseSchema,
   configShellListDetectedRequestSchema,
   configShellListDetectedResponseSchema,
+  configShellListDetectedResponseSchemaV10,
   configShellProbeRequestSchema,
   configShellProbeResponseSchema,
   configShellRemoveRequestSchema,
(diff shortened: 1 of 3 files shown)


hdkshingala:
@coderabbitai approve

PR #1427, opened 2026-08-25
fix(protocol): freeze config.shell.listDetected@1.0 and open 1.1 for wslHealth

pranshugupta54:
## Summary

The Protocol Compatibility gate has been red on `main` since eed54b3cf (#1378):

```
[BREAKING] unary config.shell.listDetected@1.0 response
  at properties.shells.items.properties.wslHealth
  new property added (tolerated via optional/.catch()/default) at a released
  version on a host→client slot
```

#1378 added `wslHealth` **in place** on `config.shell.listDetected@1.0` — an
already-released line. That is a same-version wire-shape change on a host→client
slot: a released peer's payload never carries the key, so its presence is a
runtime coin flip rather than a version fact. This is precisely the
`providers.list` #258 shape the gate was built to block, and `.optional()`
tolerance is parse-time hardening, not a versioning mechanism.

**This PR freezes 1.0 at the shape that shipped and opens 1.1 carrying the
annotation.** No behavior change: `wslHealth` still rides the wire for any peer
that negotiates 1.1.

| File | Change |
| --- | --- |
| `host/config/schemas.ts` | `configDetectedShellSchemaV10` / `configShellListDetectedResponseSchemaV10` — the frozen pre-probe shape; the live schema now `.extend()`s it with `wslHealth` |
| `host/config/contracts.ts` | `configShellListDetectedV10` binds the frozen response; new `V11` binds live; identity `UpgradeV10ToV11` |
| `host/registry.ts` | `latestMinor: 1`, minor `1` installed |

## Why this remedy and not the alternatives

**Not `compat-exceptions.json`.** That file's own `$comment` scopes it to
host→client paths that are **catalog-gated** and states it is "Not a per-change
review mechanism." Nothing negotiated gates `wslHealth`'s emission — it is a
plain new key on a released response, not an id behind a catalog the client
already fetched. The existing optional-property entries there
(`**.workspaceMode`, `**.activeProcessName`, `**.ackCreditSupported`) are static
policy over shapes whose absence is a *defined protocol state*; granting an
entry here to unblock one PR is exactly how the #258 hole reopens.

**Not a new major, despite the hint text.** The finding and
`released-baseline-compat.test.ts` both say "open a new major with downgrade
bridges (amp `003d7586` / devin `407d110` template)". That is right for the
growth those templates had in mind — ids/enums on a catalog payload — and wrong
here: `versioned-rpc.ts`'s cross-major check **rejects** a purely additive major
(`"Major bump 1 -> 2 for method '…' is not a breaking change (could have shipped
as a minor)"`), so a `2.0` would fail at registry load. One optional field is not
a breaking change.

This is the same correction already recorded in-tree for `providers.list@7.0 →
7.1` (`frozen-catalog-lines.test.ts:86-89`, `registry.ts:1711-1717`): *"the new
line is a MINOR … two optional fields are not one. Additive growth on a frozen
head takes the next minor."*

**No hand-written downgrade bridge is needed.** The skew is *within* a major, so
`downgradePathsFromLatest` (which only spans majors) stays empty. A peer that
negotiated 1.0 gets the projection the framework already performs — the newer
peer re-parses its latest-shape payload through the older installed schema
(`json-schema-fingerprint.ts`'s "frozen response reparse"), and the frozen 1.0
object strips `wslHealth`. That is also why the addition is legal as a minor at
all: field addition is additive under both `lenient` and `no-value-growth`; only
enum/union *value* growth is strict on unary responses.

## Verification

- `bun run --filter @traycer/protocol test src/host/__tests__/released-baseline-compat.test.ts` — **1/1 pass** (was the failing gate).
- Adjacent registry/versioning + config suites, 19 files / **305 tests pass**:
  `released-surface-compat`, `two-sided-release-invariant`, `released-floor`,
  `frozen-catalog-lines`, `src/config/__tests__`. The registry validator runs on
  import, so `latestMinor`/upgrade-path structure is checked by every one of them.
- `bun run --filter @traycer/protocol compile` and `lint` — clean.
- Nothing outside `protocol/` is touched; the live `ConfigDetectedShell` type is
  unchanged, so the host resolver, GUI picker and CLI printer compile as-is.

Deliberately not run locally: the full monorepo suite — CI owns it.

## Checklist

- [x] Pre-commit static checks pass
- [x] Separate CI test checks pass — see Verification
- [x] Tests added/updated where it makes sense (the existing gate is the test; it goes green)

### DIFF
--- a/protocol/src/host/config/contracts.ts
+++ b/protocol/src/host/config/contracts.ts
@@ -1,4 +1,7 @@
-import { defineRpcContract } from "@traycer/protocol/framework/index";
+import {
+  defineRpcContract,
+  defineUpgradePath,
+} from "@traycer/protocol/framework/index";
 import {
   configEnvDeleteRequestSchema,
   configEnvDeleteResponseSchema,
@@ -16,6 +19,7 @@ import {
   configShellGetResponseSchema,
   configShellListDetectedRequestSchema,
   configShellListDetectedResponseSchema,
+  configShellListDetectedResponseSchemaV10,
   configShellProbeRequestSchema,
   configShellProbeResponseSchema,
   configShellRemoveRequestSchema,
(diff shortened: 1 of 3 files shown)


pranshugupta54:
Superseded by #1426 — Tanveer's fix addresses the same post-#1378 baseline failure at the root (the cli-v1.2.0 release-tag race) and is green. Closing to keep one lane.

PR #1428, opened 2026-08-25
fix(gui-app): treat a disabled host query as waiting, not empty

pranshugupta54:
## A deep link to a line silently loses its jump

Open a workspace file at a line — a `file.ts:412` link, a find result, a diff jump. If the tile mounts before its tab host client has resolved, this happens:

1. `useWorkspaceReadFile` is disabled, because `useHostQuery` disables itself whenever the client is null or host readiness has not landed.
2. A disabled TanStack query is `status: "pending"` with `fetchStatus: "idle"`. `isLoading` is defined as `isPending && isFetching`, so **`isLoading` is `false` for a disabled query.**
3. `isSettledUnconsumableWorkspaceFile({isLoading: false, hasError: false, hasContent: false})` returns `true` — the read is judged *settled and unconsumable*.
4. The effect calls `clearWorkspaceFileRevealTarget`. The pending line target is gone.

Then the client resolves, the file loads normally, and the user is at line 1.

This is **silent and unrecoverable**. Nothing looks broken — the file is right there, fully rendered. The eviction is one-way and the target is gone for the tile's life, so there is no second chance once content arrives. The user just quietly ends up somewhere they did not ask to be, with no reason to suspect the link did anything other than work.

The same `false` drives two more misreports in the same tile: the preview body skips its spinner and returns `null` (blank pane, no spinner, no message), and the find bar reports **"File content is unavailable."** instead of "File is still loading."

## The window is ordinary, not exotic

A reviewer will reasonably assume "null host client" is a narrow edge. It is wider than that. `useReactiveHostReadiness` requires **both** `getActiveHostId()` and `getRequestContextUserId()` to be non-null, and a pinned requester answers `null` for `getActiveHostId()` until the host's directory row lands — its own doc says so. So a **non-null** client is still `!isReady`, and the query is still disabled, on an ordinary cold mount. Add a slow host, a host re-point, or a tab whose host is remote, and this is routine rather than rare.

## Someone already hit this and had no way to fix it properly

`file-change-segment.tsx` carried this comment, which this PR deletes:

```ts
// isLoading (not isPending): a disabled query (e.g. both hashes null) keeps
// isPending true forever; only show the spinner while genuinely fetching, and
// otherwise fall through to the reason copy.
```

That is the bug reasoned about correctly and then resolved the wrong way — and for a good reason. `isPending` is right for the host-disabled case and wrong for the caller-disabled case (both hashes null), and nothing available at that call site could tell the two apart. `isLoading` was the only flag that did not hang the surface on a spinner forever, so `isLoading` is what shipped.

That is the argument for this helper existing. It is not an abstraction looking for a use; it is the missing distinction that forced a wrong choice.

## This gate is already the house pattern

One surface had worked the trap out and encoded the same gate by hand. `chat/composer/picker/use-mention-items.ts:165-177`:

```ts
// "Requested" mirrors the query's real enable condition, including the
// client: with no hostClient the query is disabled and no rows are ever
// coming, so the zero-match verdict must not wait on it (a disabled query
// pends forever - gating on isPending would pin the menu open offline).
const terminalsRequested =
  active && currentEpicId !== null && hostClient !== null;
```

That is `requested` — same name, same reasoning, same conclusion, reached independently. This PR does not introduce a pattern; it makes the one already in the codebase reusable, so the next surface does not have to rediscover it or, as in `file-change-segment`, settle for the flag that fails the other way.

## The fix

```ts
export function isHostQueryAwaitingData(args: {
  readonly query: HostQueryArrivalState;   // { readonly isPending: boolean }
  readonly requested: boolean;
}): boolean {
  if (!args.requested) return false;
  return args.query.isPending;
}
```

`requested` is the caller's **own** gate — the condition it passed as `enabled`, or `true` if it passed none. It is what separates the two ways a query can end up disabled: a query with no host to ask is *waiting*; a query the caller switched off awaits nothing. Without it the predicate collapses to `isPending` and spins forever on the caller-disabled case — exactly the trap the deleted comment describes.

`requested` is the site's real gate at all three call sites, not `true` by assumption:

| Site | `requested` |
| --- | --- |
| `workspace-file-tile.tsx` | `hasWorkspaceReadFileTarget(node.workspacePath, node.filePath)` |
| `snapshot-hash-inline-diff.tsx` | `hasSnapshotDiffContentToRead(props)` |
| `file-change-segment.tsx` (live arm) | `hasSnapshotDiffContentToRead(props.segment)` |

Each of those is the same function the corresponding hook uses **as** its `enabled`, so the call site and the gate cannot drift.

### Why four new files for a three-site fix

The two gate predicates started out beside their hooks, which is where single-source-of-truth wants them. That breaks 11 test files: they `vi.mock` `@/hooks/workspace/use-read-file-query` and `@/hooks/snapshots/use-snapshot-diff-query` with factories exporting only the hook, so a second export from either module resolves to `undefined` at runtime. Hence `lib/workspace/workspace-file-read-target.ts` and `lib/diff/snapshot-diff-request.ts` — each hook imports its own gate from there and uses it as its `enabled`, so there is still exactly one definition, no mock churn, and a pure predicate ends up in `lib/` rather than in a hooks module.

## Sites fixed

**`workspace-file-tile.tsx`** — one honest value now feeds all three consumers: the reveal-target eviction, the preview body's spinner, and the find environment. Renames `WorkspaceFilePreviewContent.isLoading` and `WorkspaceFileFindEnvironment.isLoading` to `awaitingContent`, and `isSettledUnconsumableWorkspaceFile`'s `isLoading` argument likewise.

**`snapshot-hash-inline-diff.tsx`** — was rendering `FILE_EDIT_REASON_COPY["blob_missing"]`, i.e. **"Skipped — snapshot blob missing."**, while the tab host client resolved. A hard verdict about the owner's snapshot store, printed for a read that had not been attempted.

**`file-change-segment.tsx`** (live arm) — same wrong copy, same cause. `FileChangeDiffView.isLoading` becomes `awaitingContent`. The **published** arm is not host-query-backed: `usePublishedSnapshotDiff` computes the flag itself in `snapshotDiffResult` (`published-chat-source.tsx:150`), where it is already `true` for a side pending-without-error and `false` for a read never asked for. That is the awaiting-an-answer fact rather than a TanStack flag, so it passes through unchanged under the honest name, with a comment at the call site saying so.

## Test

`lib/query/__tests__/host-query-awaiting-data.test.tsx` drives the disabled case through a **real `useHostQuery` with `client: null`**, not a hand-built result object — believing the wrong thing about TanStack's disabled semantics is what made this bug possible, so the test pins the actual behaviour. It asserts `status === "pending"`, `fetchStatus === "idle"`, **`isLoading === false`**, `data === undefined`, and only then that the helper reports `true`.

A test that only covered pending-and-fetching would pass on the broken code. This one fails on it. The caller-disabled case is in for the opposite reason: it is what stops a later "simplification" back to bare `isPending`.

## Why nobody noticed: the fixtures described an impossible state

The tile's own find tests modelled a query result with `isLoading` but no `isPending`, and included this case:

```ts
function missingReadFile(): ReadFileState {
  return { data: undefined, isLoading: false, isError: false, error: null };
}
// asserted: coverageMessage === "File content is unavailable."
```

**`data: undefined` on a settled, non-error query is not a reachable TanStack state.** A query holding no data is either pending or errored; a successful one always has data. So that fixture was describing the *disabled* case — and asserting it renders "File content is unavailable."

The empty verdict wasn't merely unnoticed. It was encoded as expected behaviour and defended by a green test, which is why it survived: anyone who looked had a passing suite telling them the empty verdict was right. The general form is worth stating plainly — **when a fixture describes an impossible state, the assertion built on it is describing the bug.**

The repair keeps three genuinely reachable cases rather than flattening them:

| Fixture | State | Renders |
| --- | --- | --- |
| `loadingReadFile` | pending, fetching | "File is still loading." |
| `disabledReadFile` *(new)* | pending, idle — no host bound yet | "File is still loading." |
| `missingReadFile` *(corrected)* | settled payload, `content: null` | "File content is unavailable." |

So the disabled case is now pinned at the tile level too, not only in the helper's unit test.

## Surveyed, deliberately not fixed

There are ~75 `.isLoading` reads under `clients/gui-app/src`. Only the three above are fixed here.

**`chat-usage-dialog.tsx:136` and `epic-usage-dialog.tsx:250` — a real instance, deliberately held.** Both take `useUsageSummaryForClient(useHostClientForHostId(target.hostId), …)` with a nullable client; disabled means skipping `UsageDialogSkeleton`, `error` null and `data` undefined, so both fall through to `UsageDialogUnavailable` — "Usage data unavailable." — while the client is still resolving. It was independently re-found by a second reviewer during this work, so the omission is known and deliberate rather than missed. It is held back because "unavailable" is a plausible *real* answer for a host that does not report usage, so getting these right needs a judgment about that surface's own semantics rather than a mechanical application of the helper — and two files answering a different question is how a tight PR becomes a debatable one. One line each as a follow-up now that the helper exists.

**`lib/commands/sources/open/terminals-subpage.ts:410-414`** — `useTerminalList({kind: "epic", epicId: ctx.activeEpicId ?? ""}, hostClient)` with a nullable `useHostClientForHostId(activeEpicHostId)`. Disabled renders an empty terminal list in the command palette rather than a loading state. Same class, found during this work, left alone to keep scope at three; a follow-up alongside the usage dialogs.

Others, with reasons:

- `comments/comment-sidebar.tsx:96` — **already handles it**, deliberately: `isUnavailable = data === undefined && fetchStatus !== "fetching"`, with a comment naming the disabled-cold-query case. Routing it to "unavailable" rather than a spinner is a documented product choice.
- `chat/segments/autonomous-resume-segment.tsx:314` — hand-rolls `isPending || (isFetching && data === undefined)`, already correct for the host-disabled case. It has the *opposite* latent bug (spins forever if its `enabled` prop is false); a different defect, not this one.
- `epic-canvas/sidebar/epic-sidebar-file-tree.tsx:292` — disabled renders a blank tree with no spinner. Real, but benign: no wrong verdict and no destructive state.
- `layout/dialogs/desktop/open-epic-in-new-window-dialog.tsx:244` — gated by `isAvailable` ahead of `isLoading`, and not `useHostQuery`-backed.
- `usage-analytics/usage-summary-panel.tsx:493`, the `settings/**` host-scope surfaces (`use-host-options`, `host-switcher`, `host-scope-gate`, `add-host-dialog`), `home/host-workspace-selector/**`, `pickers/harness-model-picker.tsx`, `hooks/composer/use-workspace-entries.ts`, `hooks/harnesses/use-gui-harness-catalog.ts` — these aggregate `isLoading` across several queries into a list-is-loading flag whose empty state is "no hosts / no workspaces yet". The host-scope surfaces in particular exist precisely to run before a host is bound, so their gating is a separate design, not this bug.
- `epic-canvas/mobile/switcher-terminals-list.tsx` and `sidebar/use-epic-terminals-panel.ts` — owned by another PR, untouched.

## Notes on the run history

A reviewer scrolling the checks will see a force-push and two earlier red runs. Both are accounted for:

- **`protocol-compat` / `test (@traycer/protocol)` red on the first head — inherited, not introduced.** The blocking finding was `config.shell.listDetected@1.0` → `wslHealth`, added by `eed54b3cf` (#1378), the commit this branch forked from; `Protocol Compatibility` was already failing on `main` itself at that commit. Main had since landed the fix (`dbbe421e7`, #1426), so the branch was rebased onto it. That is the force-push. Diff versus `origin/main` is unchanged by the rebase.
- **`test (desktop darwin + packaging)` red once on the second head — not reproduced.** `host-lifecycle.test.ts:972`, `expected 'available' to be 'busy'`, on a probe-degradation ladder. No path exists from a gui-app-only diff to `electron-main/host`, the job is green on `main` at the rebase base and the two commits before it, and it passed on the next run. Signature recorded in the epic's standing-flakes register.

Formatting is not where this will fail: `prettier --write` reports all touched files unchanged.


pranshugupta54:
Closed by decision: the guarded window (host client pending while a file tile is open) was never reproduced on a device, and the fix as written trades the false error for an unbounded loading state on a genuinely unreachable host. Branch preserved; resurrect only if the symptom (lost line-jump / false 'unavailable' flash during reconnect) is actually reported.

PR #1433, opened 2026-08-25
fix(gui-app): reach every control the touch-target contract already covers

pranshugupta54:
## The contract, and why it was only half kept

Three stylesheets grant a >=44px coarse-pointer hit area, and all three do it the
same way — a descendant selector rooted at a scope attribute, keyed on the
shadcn **slot** attribute:

| Stylesheet | Scope | Slots granted |
| --- | --- | --- |
| `settings/settings-touch-targets.css` | `data-settings-touch-scope` | `switch`, `button`, `select-trigger` |
| `home/home-touch-targets.css` | `data-home-touch-scope` | `button`, `select-trigger` |
| `layout/shell/mobile-shell-touch-targets.css` | `data-mobile-shell-touch-scope` | `button`, `tabs-trigger` |

So a control is enlarged only if it is **(a)** inside a scope and **(b)** emits a
granted slot. Neither half of that names the other, so both halves type-check,
render, and review cleanly while covering nothing. The app broke the contract in
three distinct ways, and this PR closes all three.

Note the root font is **15px**, so every rem renders at 0.9375x — `size-7`
paints 26.25px, not 28. Nothing here resizes a control; the fix is slot and
scope coverage, and every change is invisible at every pointer type.

### 1. Right scope, wrong element type

Hand-rolled `<button>`s inside a correct scope emitted no slot, so the rules
never matched them.

| Control | File | Painted |
| --- | --- | --- |
| Notification mark-read | `notifications/notification-row.tsx` | 18.75px |
| Composer attach / mic / service tier / permissions / harness-model | `home/toolbar/toolbar-buttons.tsx` | 30px |
| Theme mode segments | `settings/controls/theme-mode-toggle.tsx` | 26.25px |
| Theme preset / font pickers | `settings/controls/{theme-preset-picker,font-picker}.tsx` | ~30px |
| Terminal cursor-style segments | `settings/controls/terminal-cursor-style-picker.tsx` | 30px |
| Settings back-crumb | `layout/header/mobile-app-header.tsx` | 37.5px |

Each gets `data-slot="button"`. Two notes on the choices:

- **Fixed at the primitive, not the call site.** `ToolbarIconButton` and
  `ToolbarPillButton` are one choke point covering six call-site files, so the
  composer toolbar is two lines rather than a dozen.
- **The crumb is a `<Link>`, i.e. an `<a>`.** That is not a hack: shadcn's own
  `Button asChild` renders an anchor carrying `data-slot="button"`, so this is
  the primitive's existing semantics rather than a new convention.

The slop is vertical-only in every scope, which is what the two segmented
controls need — their segments are flush **horizontally**, so horizontal slop
would let each capture its neighbour's taps.

### 2. Right element type, unreachable scope

`ui/dialog.tsx` and `ui/sheet.tsx` render a real `Button size="icon-sm"` close X
(26.25px), but their content is **portalled outside `#root`**. No ancestor
carries a scope, so a descendant selector can never reach it — this affected
every dialog and sheet that doesn't override `showCloseButton`.

The mechanism is a primitive-level hit area, per the rule the codebase already
states in `settings/SETTINGS.md`: *"Anything living in a portal owns its own
target instead, at the primitive."* New `ui/coarse-pointer-hit-slop.ts` exports
`PORTAL_ICON_BUTTON_HIT_SLOP`, consumed by both primitives.

Three decisions worth surfacing:

- **A pseudo, not a bigger box.** `pointer-coarse:size-11` was rejected: the
  `ghost` variant paints on hover and while expanded, so growing the box is a
  visual change. The pseudo is a pure hit area and never paints.
- **It rides `::before`, deliberately.** An element has a single `::after`, and
  the scope stylesheets claim it. A portalled surface can still set a scope on
  its own content — `remote-folder-picker-dialog.tsx` puts
  `data-home-touch-scope` on a `DialogContent` — so a close X *can* be in reach
  of both rules at once. `::before` cannot collide. The test asserts the
  constant is `before:` and **not** `after:`, so this survives the PR.
- **Symmetric slop, because it is a corner control with no flush neighbour.**
  `-inset-2.5` is 0.625rem a side, taking `size-7` to 45px on both axes.

### 3. Missing scope entirely

`stores/tabs/use-system-tab-modal.ts` routes phones to the `/epics` **page**,
but `data-home-touch-scope` was only on the **modal** host. `history-surface.tsx`
carried no scope at all — so the entire History chrome (Sort, Filter, Select,
Refresh, Sweep, Delete, Clear filters) was unslopped on the one path a phone
actually takes. Those controls are all already slotted `Button`s, so adding the
scope to `HistorySurface` is the whole fix.

### Flush-stacked rows are sized, not slopped

`epics/epics-filter-popover.tsx`'s `FilterOption` is a raw
`<button role="checkbox">` that stacks with no gap, inside a portalled popover.
Invisible slop would overlap its neighbours, so the row grows itself —
`pointer-coarse:min-h-[44px]`.

### Every size here is a pixel literal, because rem is elastic

`--ui-font-size` is a **user setting**. `providers/theme-provider.tsx` writes it
straight onto the root element, and `stores/settings/settings-store.ts` clamps it
to **10–20px**. So every `rem` in the app scales with it, and the "15px root"
figure is a *default*, not a constant.

That makes the common `pointer-coarse:min-h-11` idiom (2.75rem) wrong for a touch
target: it is 44px only above a ~16px root, and **27.5px at the 10px minimum** —
the smallest targets for the users who chose the smallest text, which inverts the
intent. The same applies to a rem-based pseudo inset.

Both sizes this PR introduces are therefore pixel-floored:
`pointer-coarse:min-h-[44px]` for the row, and `size-[max(100%,44px)]` for the
portal pseudo — the latter matching what the three stylesheets already do
(`max(100%, 44px)`), so it holds the floor at any root size and still grows with
a larger control.

The test asserts this as an invariant rather than pinning a token: it requires
`44px` to be present **and** rejects the rem-scaled forms outright, so the idiom
cannot come back through a later edit.

## The test

`src/__tests__/touch-target-slot-contract.test.tsx` parses all three
stylesheets and checks the components against what the CSS **actually** grants,
rather than restating the list. A test that pinned one component's class string
would pass forever while the next raw `<button>` regressed the same contract.

- **Set equality between the two rules, per file.** The slop pseudo is
  `position: absolute`, so it resolves against the nearest *positioned*
  ancestor. A slot added to the `::after` rule but not the `position: relative`
  rule still gets a pseudo — one hung off whatever ancestor happens to be
  positioned, silently placing the hit area somewhere other than on the control,
  with nothing visible to catch it in review. The test unions across rules in
  each direction, which tolerates settings' split shape (its switch has its own
  `::after` rule) while still failing a slot that appears in only one.
- **The `.absolute` asymmetry is asserted as asymmetry.** Home and mobile shell
  exempt `.absolute` controls from the `position: relative` fallback; settings
  does not, because it contains no such control. The test records the expected
  value per scope, so it pins what is there instead of a uniformity the app does
  not have.
- **44 is asserted as a pixel literal**, with the 15px-root reason inline — the
  `min-h-11` idiom is 41.25px and leaves a control short.
- Six primitives are **rendered** and asserted to emit a slot in their scope's
  granted set — including `ThemePresetPicker` and `FontPicker`, which turned out
  to mount standalone with no harness, so neither needed downgrading. Three more
  are **source-pinned**, and the doc comment says
  plainly that this is the weaker half: it proves the attribute exists, not
  where it lands. Those three live in surfaces that need a feed row, the router,
  or the composer to mount.
- `epics/__tests__/history-surface.test.tsx` is **extended** rather than
  paralleled: it asserts the list is a *descendant* of the scope, since a
  descendant selector is the actual mechanism.

## Two changes a reviewer should not read as noise

- **`home-touch-targets.css` header comment.** It enumerated its scope-setters
  ("set by HomePage and by the system-tab modal host"), and adding a third made
  that sentence false. Rewritten to state the invariant instead of a list.
- **`cleanup()` added to `history-surface.test.tsx`'s `afterEach`.** The suite
  runs `globals: false`, so Testing Library registers no automatic teardown;
  that file survived only because it had a single `render`. The second test
  added here would have found the first test's tree still mounted and thrown on
  an ambiguous query. Latent bug in a shared harness, one line, fixed in a file
  this PR was already extending.

## Follow-ups, with evidence

Each of these needs layout or labelling work rather than a slot, so it is
deliberately **not** in this PR.

**Editor / mermaid / wireframe toolbars are outside every scope.** This one is a
design decision rather than a deferred edit, so it is written up on its own —
mechanism, both options with their honest costs, and a recommendation — in the
`editor-toolbar-touch-targets` spec artifact. Summary below.
`editor-core/toolbar/toolbar-button.tsx` renders a bare `<button>` sized to
26.25px by `editor-core/styles/editor.css`. It reaches the screen through
`artifact-toolbar.tsx` -> `epic-canvas/renderers/collab-tile-body.tsx` ->
`ActiveTabBody`, inside `epic-canvas/mobile/mobile-epic-tile-view.tsx` — which
carries **no** scope attribute (the only one in that file is on the empty-pane
branch). So adding a slot there would be a silent no-op, and the nearest
scopeable ancestor is the whole tile body. Two options, and picking between them
is a real decision rather than an edit:

1. Scope the tile body — one attribute, but it hands slop to every button in
   the chat, terminal and diff surfaces at once.
2. A `@media (pointer: coarse)` block in `editor.css` on
   `.tc-editor-toolbar-button`, which already owns the height. It is the single
   choke point for all four toolbars and needs no scope, so it also covers the
   portalled fullscreen wireframe dialog — but it is a **fourth** slop
   mechanism, in a package documented as design-token-neutral, and this PR's
   thesis is one contract across three.

Two findings from writing that up, both of which change the shape of the fix:
the class `.tc-editor-toolbar-button` is a **wider** choke point than the
`ToolbarButton` component (the block error boundary styles a raw `<button>`
directly, so a component-level fix misses it); and the mermaid/wireframe block
toolbars are `opacity: 0; pointer-events: none` until `:hover`, which touch
devices do not have. For those two surfaces, being summonable is upstream of
being tappable — and `editor.css` carries no coarse-pointer rules at all today.

**`settings/host-scope/host-switcher.tsx` "Try again" / "Add host…"** (~22.5px).
`HostSwitcher` has ten call sites and most are themselves popovers or dialogs,
so a slot would fix the settings-sidebar instance and silently miss nine. At
`text-ui-xs px-1 py-0.5 self-start` these are link-styled affordances; making
them tappable everywhere is a primitive-sizing decision.

**`layout/shell/mobile-nav-drawer.tsx` "View all"**. It sits in a `sticky py-1`
header roughly 28px tall, directly above the task rows — so 44px of slop would
spill about 8px onto the first row, every one of which is a nav target. That is
the flush-stacked hazard; it needs the row-sizing treatment, not `::after`.

**Visible labels for the terminal cursor-style picker.** Correcting the record:
the three glyph buttons each carry a real `aria-label`, so the tooltip is the
*visual* label, not the accessible one — which is why the mechanical half was
safe to keep here. Giving them visible labels is a redesign.

## Note for whoever merges second

An opt-out attribute mechanism (`data-touch-slop-opt-out`) for flush-stacked
rows arrives with the mobile parity wave. This branch uses the idiom that exists
here today — `pointer-coarse:min-h-11` at the primitive, per `ui/select.tsx` and
`ui/dropdown-menu.tsx`. The two will meet on merge. They are compatible, but
please reconcile them rather than inventing a third.

**This PR does not conflict with the comments-lane PR, in either merge order.**
That is worth stating because it looks like it should: both touch the same
`:is()` slot lists. The contract test here reads the granted slot set **off the
stylesheets at runtime** rather than restating it, so a slot another branch adds
is picked up automatically and needs no edit here. The test fails only if a slot
lands in one of the two rules and not the other — which is the trap it exists to
catch, not a collision between branches.


pranshugupta54:
Closed by decision: no user-reported symptom behind the enumerated controls; branch preserved if any of them ever gets reported.

PR #1434, opened 2026-08-25
fix(gui-app): give the chat composer the bottom safe-area gutter

pranshugupta54:
## The bug

The chat tile mounts its lower-surface stack as `pointer-events-none absolute inset-x-0 bottom-0 z-10` (`chat-tile.tsx`), so the stack's bottom edge is the tile's. On a phone the chat tile renders through `MobileEpicTileView`, whose `paddingBottom: keyboardInset` applies only to terminal tiles — so for a chat, tile bottom is the viewport bottom.

Inside the stack the order is approval surface → lower dock → composer region, and the composer is last. Its outer `[data-chat-composer]` edge lane carries no vertical padding, so the centered backplate's `pb` is the only thing between the toolbar row and the edge of the screen.

That was a bare `pb-4` (1rem). `#root` reserves the top and both horizontal insets app-wide but deliberately not the bottom — per `AGENTS.md`, the bottom edge is what is left to a surface. The iPhone home indicator's inset is ~34px, so the send button and toolbar row sat inside its strip. Every other phone-bottom surface in the app already takes the token; the most-used one did not.

## The fix

One class on the composer backplate:

```diff
-... max-w-3xl bg-canvas pb-4 after:...
+... max-w-3xl bg-canvas pb-safe-bottom-gutter after:...
```

## Which token, and why

Two exist in `index.css`:

| token | value |
| --- | --- |
| `pb-safe-bottom` | `var(--safe-area-inset-bottom)` |
| `pb-safe-bottom-gutter` | `max(1rem, var(--safe-area-inset-bottom))` |

**`pb-safe-bottom-gutter`.** The bare inset is 0 on every desktop, every browser on a laptop, and every phone without a home indicator — so a straight `pb-4` → `pb-safe-bottom` swap would have *removed* the composer's existing 1rem of breathing room on all of them, trading the overlap for a visual regression everywhere else. The gutter keeps that 1rem where the inset is zero and grows to the inset where there is one. `landing-draft-surface.tsx` and `composer-options-sheet.tsx` use the gutter form for the same reason.

`bg-canvas` sits on the same box, so the padding is painted rather than left showing transcript through, and the 1px `after:` seam seal rides down with the padding box.

`tailwind-merge` already knows the token — `safe-bottom-gutter` is in `SAFE_AREA_SPACING_TOKENS` and wired into the `pb` class group in `lib/utils.ts`, so it conflicts with (rather than coexists with) any `pb-*` an override brings. Moot for this edit, which replaces the token inside the same string literal, but the guarantee holds for the next caller.

## `ComposerSlotShell` deliberately untouched

`chat-tile-lower-surfaces.tsx` carries the same `pb-4`, and it is **not** always the bottom-most surface:

- the approval queue renders it *above* the lower dock and the composer;
- the unanswerable-interview escape hatch renders it *above* the composer, with the next surface taking `pt-0` to connect flush.

In both, the `pb-4` is inter-surface spacing. Padding the shared shell would have inserted a ~34px gap in the middle of the stack on a phone — a visible regression in two states, shipped as a fix.

## Named follow-up: the composer region's two other bottom-most states

`ComposerSlotShell` *is* the bottom-most surface in the two branches that stand in for the composer, and both have exactly the same overlap this PR fixes, in a different state:

| Call site | State |
| --- | --- |
| `chat-tile-lower-surfaces.tsx:524` | read-only viewer notice |
| `chat-tile-lower-surfaces.tsx:558` | pending interview card |

They are not in this diff because the shell's `pb-4` is load-bearing for its non-bottom-most uses above, so the fix cannot be the same one-class swap. The shape it wants is a third `ComposerSlotBottomSpacing` variant — `"normal" | "none"` gains a safe-area case that the bottom-most call sites pass and the stacked ones do not. That is a change to a shared component's API, which is why it is a separate PR rather than a widened scope here.

## Test

Extended `safe-area-contract.test.ts` with a per-class-list pin on the composer backplate: it must contain `pb-safe-bottom-gutter` and must not carry a bare `pb-<number>`. Pinned per literal rather than per file, because the same component paints a rate-limit banner whose spacing would otherwise vouch for the backplate's. Asserts the contract, not a pixel value.

The existing source-shape pin in `chat-scrollbar-composer-overlay.test.tsx` was updated to match.

## Verification

- `bunx vitest run` on both touched test files — 24 passed.
- `bun run compile` (`tsc -b`) in `clients/gui-app` — exit 0.
- `bunx eslint` on the three changed files — clean.
- `bunx prettier --check` on the three changed files — clean.


pranshugupta54:
Closed by decision after device review: the finding measured the composer's own bottom padding in isolation, but on the assembled phone screen the composer is not the bottom-most element — the toolbar/tile-bar row below it provides the home-indicator clearance, and the landing composer's surface already carries pb-safe-bottom. No user-visible defect. Branch preserved.

PR #1921, opened 2026-09-15
gui-app: Home says running, cold tasks stay one row, agent glyphs take icon colours, Home tab shares the tab chrome, one background mark

pranshugupta54:
Five small polish items on the Home tab and the composer's compact chips.

- **Home says `running`.** The status cell's mid-turn word was the wire term `turn`; it now prints `running` (`● running` / `○ background`) everywhere Home spells the tier, including the cold-task summary and Stop-all lists. The state keys and the wire fields are unchanged.
- **Cold tasks stay one row.** A task that is not open in this window used to nest a placeholder `Agent` row per running agent, because the activity feed carries ids but no names. It now renders as a single row with its summary and `Stop all` until the task is open here and the names are known; prompts and jobs the window does know about still hang off the task.
- **Agent glyphs take the icon colours.** Home's chat and terminal-agent glyphs now honour Settings ▸ Appearance's per-type icon colours through the same rule every other surface uses (extracted as `useEpicNodeIconTone`); the task glyph stays muted since no task colour exists.
- **Home tab shares the tab chrome.** The Home item used a fixed box, so its bubble was narrower and taller than its neighbours; it now uses the shared header-tab classes with the same height and horizontal padding.
- **One mark for background work.** The composer's compact background chip uses the notifications' `MessageSquareClock` glyph for every kind of background item instead of a per-kind icon or a stacked-layers fallback; shimmer, count and tooltip are unchanged.

SETTINGS.md's Home disclosure paragraph is brought up to date with the no-auto-expand rule.

## Testing

Home, tab-strip, canvas icon, compact-chip and background-tree suites, plus lint, `oxfmt --check` and compile.

Signed-off-by: Pranshu Gupta <pranshu@traycer.ai>


pranshugupta54:
@codex review

PR #1933, opened 2026-09-16
feat: open a remote shell's output window from its chip

tanveergill:
## Summary

A shell started on another host by an agent (`traycer_run_shell` with `host`) now delivers its notifications to the creating agent's host (internal traycer-internal#5686). The chip that renders such a delivery in the chat transcript, and the queued-delivery badge before dispatch, open the output window through the tab's host, where a remote shell's log does not exist.

The output window already streams from whichever host its tile names. This change carries the shell's host on the chip and hands it to the door:

- `managedCommand.hostId: z.string().nullable().default(null)` on the persisted resume trigger and on the queued managed-command item (same defaulted-key pattern as `monitoring`: an older host strips it on parse instead of failing the chat). Null keeps today's meaning, the tab's host. Compat exceptions added on the same terms as the existing `managedCommand` exemption; schema surface baselines regenerated (additive).
- `useManagedCommandDoor` takes `(commandId, hostId | null)` and opens on `hostId ?? tabHostId`; the divider and the queued badge pass the chip's host, the three local callers pass null.
- The divider's presence gate treats a foreign-host shell as `unknown` rather than `gone`, so the door stays open; the window gives its own account. A remote shell's tab title falls back to "Output" since no chat stream on this canvas holds its record.

## Test plan

- protocol: released-baseline compat, surface compat, schema-surface compat, content-blocks and chat-subscribe suites green.
- gui-app: door tests for a divider with and without `hostId`, chat-surface tests, fixtures updated; `compile` green.


tanveergill:
@coderabbitai full review

tanveergill:
@codex review

PR #1934, opened 2026-09-16
[probe, do not merge] trace hidden->visible replay drift in chat-messages test

tanveergill:
Temporary instrumentation to capture CI-only behavior of the chat-messages hidden->visible test (fails on main since #1910 with expected 2160 to be 2200). Will be closed without merging.


tanveergill:
Probe captured; closing without merge.

PR #1956, opened 2026-09-16
docs(protocol): name 1.12, not 1.11, as the draft-image refusal cause line

hdkshingala:
Three comments in `chat-frame-compat.ts` say the typed draft-image refusal `cause` rides `chat.subscribe@1.11`. It rides `1.12`, and the gate a few lines below one of them has always said so:

```ts
if (negotiated !== null && negotiated.major === 1 && negotiated.minor >= 12) {
  return frame;
}
```

`DRAFT_IMAGE_CAUSE_MINOR = 12` agrees, and so does `subscribe.ts`'s own contract doc. So a `1.11` peer is never sent this field, and these comments — in the file whose entire job is the version projection — said that it is. The one at line 28 is the docstring of the stripping function itself.

**Why it drifted.** The draft-image work was built against `1.11`. `main` took that minor for the cross-host shell host while it was in review, so the cause moved up one. The constants and tests moved with it; prose did not, and shipped in #1917.

**Comments only.** `git diff` touches no statement — the gate, the projection and the exported surface are untouched — and the 137 tests in the two suites that own this number (`chat-frame-compat`, `chat-subscribe-line-surfaces`) pass unchanged.

The reworded docstring points at `DRAFT_IMAGE_CAUSE_MINOR` as the one place a machine checks this number, since a version named only in prose is exactly what went stale here.

Found while reading this file for the host wire in traycerai/traycer-internal#5668, where the same stale `1.11` appeared in three host-side comments — each naming a **different** minor (1.11, 1.10, 1.10). Those were corrected there. A third reference here (line 665, "the newer 1.11 acknowledgement downgrade") was missed by my first sweep of this file because it describes the projection without using the words "cause" or "image"; this PR covers all three.


hdkshingala:
@coderabbitai approve

`bca032012`. All 38 checks pass, including `test (desktop darwin + packaging)`, which flaked six times earlier today across four different tests and one OOM.

You approved `cfa9d693d`; the only change since is `bca032012`, which corrects a sentence in the same docstring. The previous wording said `DRAFT_IMAGE_CAUSE_MINOR` is "the one place that names this number for a machine to check" — untrue, since `subscribe.ts`'s contract, the gate in this file and three test files all name `12`. It now says what is actually true and useful: that constant drives an assertion across every minor from `1.0` up, so it is the one that catches a drift rather than restating the answer.

Still comments only. `git diff` touches no statement on either commit; the gate remains `negotiated.minor >= 12`.

hdkshingala:
@coderabbitai approve

`cdf519cbe`. You approved `bca032012`; the only change since is a merge of `origin/main` (three commits: #1954, #1943, #1957) to satisfy the strict up-to-date rule. Zero file overlap, no conflicts, compile clean. This PR remains comments-only — `git diff` against the merge base touches no statement.

PR #1980, opened 2026-09-18
feat(chat-search): filter by whose words - All messages, Your messages, Agent replies

tanveergill:
## Why

The Search chats author filter read **Anyone / People / Agents** and was read as "user messages vs assistant messages". It was not: it only narrowed user-turn prompts by sender (typed by a person vs sent by another agent), and assistant replies passed through under every choice.

## What changes

The filter now means *whose words are searched*:

| Label | Matches |
|---|---|
| **All messages** (default) | everything, including prompts other agents sent and notices |
| **Your messages** | only the prompts a person typed |
| **Agent replies** | only the chat's own replies |

- `protocol`: `chatSearchRoleFilterSchema` becomes `any | human | assistant` (`agent` removed). `chat.search` 1.0 has not shipped in any release tag, so the enum changes in place rather than adding a minor.
- `gui-app`: the dropdown labels and its accessible name (`Message filter`).

Host side (query clause, tool catalog, tests): traycerai/traycer-internal PR to follow, which re-pins the submodule to this squash.


tanveergill:
@codex review

PR #1992, opened 2026-09-18
fix(comm-graph): stub only the foreign end of a cross-task row, and name the listed task

tanveergill:
## Summary

Two small follow-ups from the cold review of cross-task agent-to-agent messaging.

- `gui-app` comm-graph: `commGraphPeerTaskStubs` marked **every** endpoint missing from the current agent list as a peer-task stub. If the local agent was deleted, or the record feed lagged the event feed, both ends were missing and the local agent was drawn as living in the other task, with a button that opened it. A cross-task row has exactly one foreign end, so now exactly one missing endpoint is the stub, and a row with both ends missing (or a null end) produces no stub.
- `protocol` `formatAgentListResponse`: when the listing carries the host's `epicId` enrichment, the heading reads `Agents in task '<id>' (relative to you):`, so a listing forwarded between agents is addressable. Absent the key (the CLI path, which parses through the stripping schema) the wording is unchanged. No schema change.

## Test plan

- [x] `comm-graph-model`: both-missing row yields no stub; an unplaceable row does not poison later rows. Both fail against the old loop.
- [x] `agent-list-format`: heading with and without the key; CLI legacy expectation untouched
- [ ] CI


tanveergill:
@codex review

PR #1993, opened 2026-09-18
fix(protocol): drop what a frozen line cannot represent, instead of its whole catch scope

hdkshingala:
Follow-up to #1977, which froze `providers.list@9.1` against four leaves and had to admit in its own docblock that those were "NOT every live schema this row can reach". This closes that, and along the way found that the freeze half alone would not have helped.

> **Read the last commit first if you reviewed an earlier push.** A cold read found this PR's own prose claiming more than its code does. The mechanism is real and one field it fixes is live today; most of it is **armed and dormant** by design. That distinction is now stated in the code and is corrected throughout what follows.

## The measurement that changed the design

I walked the zod graph of every `providers.list` row (alias-aware, so `PROVIDER_AUTH_SCHEMA = ..._V20` is not miscounted as live). Each released row — `@7.0`, `@8.0`, `@9.0`, `@9.1` — reaches **30 enum leaves that sit inside an array**, and **27 of them share one catch scope**: `providers[].nativeCapabilities`, a whole-object `.catch(DEFAULT)`.

So one new settings tab costs a 7.0 peer its MCP, Plugins and Skills tabs **together**.

**Seven** leaves have no `.catch()` between them and the root at all, where growth fails the entire `providers.list` response rather than degrading part of it. All seven are under `native`, and they split two ways:

| | leaves | what rescues it |
|---|---|---|
| under an array | `servers[].status`, `servers[].statusSource`, `servers[].tools[].denySources[]`, `skills[].source` | a pin — the projection can then drop the member |
| scalar | `native\|0\|3.server.status`, `native\|0\|3.server.statusSource`, `native\|1.code` | a `.catch()`. Dropping cannot help; there is no array to drop from |

Worth reading that table twice, because an earlier revision of this PR had it backwards. The same `denySources` enum appears at two paths with **different** fixes — the `servers[]` one is repairable by a pin, the single-server-arm one is not.

## Why a pin alone is not the fix

Natural experiment, with a control. `modelProviders` is a member the live tab enum has and the frozen `V70Preimage` one lacks — a growth that already happened:

```
CONTROL  live   <- baseline   strict=OK   tabs=["general","mcp"]
         frozen <- baseline   strict=OK   tabs=["general","mcp"]
TEST     live   <- grown      strict=OK   tabs=["general","mcp","modelProviders"]
         frozen <- grown      strict=FAIL tabs=["general","env","usage"]   <- DEFAULT
```

Unpinned, the host forwards the unknown member and the peer's own compiled copy of the schema discards the object. Pinned, the host's own downgrade discards it first. **Same loss** — the pin moves where the collapse happens, not whether it happens.

## What actually changes the outcome

`projectOntoFrozenLine`: before the frozen reparse, an array keeps the elements the frozen line can represent and drops the ones it cannot. That is not new policy — it is the rule `downgradeProviderCliStateListToV70` already applies to the provider list itself (`flatMap`, dropping rows the frozen state rejects), pushed down the whole tree.

One exception, added after review: **it never empties a non-empty array.** For `profiles[].rateLimitLimitedScopes`, `null` means "could not read" and `[]` means "read fine, nothing limited" — emptying it would turn "rate limited" into "not limited", a confident wrong answer. Degrading to unknown is allowed; fabricating a positive claim is not.

It walks the schema rather than reading parse issues, and the reason is the problem in miniature: **`.catch()` makes the parse succeed.** `sharedWithProviders` is `z.array(providerId).catch([])`, so an unknown id reports no issue at all and silently yields `[]`. The leaves worth repairing are exactly the ones whose failure is already being swallowed.

This was not an original idea. `provider-native-schemas.ts` already carried the finding verbatim:

> FILTER `supportedTabs` before the parse; never reparse it ... the peer loses MCP, Plugins AND Skills over a single tab id it never knew. This is the trap every closed enum above is copied to avoid.

written when `projectNativeCapabilitiesToV70Preimage` was deleted, beside a note that whoever reopens the line "has to make them again". v8.0, v9.0, v9.1 and v9.2 all opened since. It never came back.

## What arms this, and what is dormant

This is the part the earlier revisions of this PR got wrong, so it is stated precisely.

`traycer-host`'s `handler.ts` parses the resolver result against the **canonical (head)** schema and 500s **before** any downgrade runs. So a value only reaches the projection if the head already accepted it — and where a frozen row binds the same enum **object** as the head, it accepts exactly what the head accepts and the walk drops nothing. **Only a leaf pinned strictly narrower than the head is ever acted on.**

Measured, per line:

```
@7.0: 2 strictly narrower of 63 enum leaves
        providers[].providerId (21 -> 19)
        providers[].managedVersions.sharedWithProviders[] (21 -> 19)
@8.0: 2 strictly narrower of 64
@9.0: 0 strictly narrower of 64
@9.1: 0 strictly narrower of 65
```

So:

**Live today** — one field:

| scenario | before | after |
|---|---|---|
| pack shared with `antigravity`, served to an 8.0 peer | `sharedWithProviders: []` — loses `claude-code` too | `["claude-code"]` |
| same, served to a 7.0 peer (lacks `reasonix` **and** `antigravity`) | `[]` | `["claude-code"]` |
| a post-v7.0 provider id on a row | row dropped | row still dropped — unchanged, and asserted |

**Armed and dormant** — everything else, including the `supportedTabs` and `denySources` cases above. The four pins in this PR are byte-identical to live *by design*, so they arm the day the live enum moves and not before. That is the intended shape of a freeze, not a gap — but it is not what "this PR fixes the deny-source crash" would have meant, and an earlier revision implied exactly that.

A test asserts that set rather than leaving a reader to assume it, and it fails the moment a pin makes a new leaf narrower.

## The pins

`nativeCapabilities` (for `supportedTabs`), `managedInstallState`, `advisory` and `autoJudge`, on every released row. Each is **byte-identical to live today** — deliberately, and that is the proof it captured the right shape: `frozen-catalog-lines.test.ts` stays green at 38/38. A pin that had captured the wrong bytes would redden it.

Worth noting one that nearly went wrong: the released baseline shows `modelProviders` **already in** the 7.0 and 8.0 response bytes, so pinning those rows against the six-member `V70Preimage` enum would not have been a freeze at all — it would have silently narrowed two shipped wires.

## The guard

`frozen-lines-reach-no-unreviewed-live-enum.test.ts` lists every live enum a released row still reaches, **by path**, and fails when the set changes.

This covers something the catalog snapshot structurally cannot: a frozen hand-copy and a live alias **dump identically**, so the snapshot stays green while a leaf silently tracks the head. Verified by ablation — reverting one pin leaves `frozen-catalog-lines` passing **38/38** while the new test reddens on `advisory.kind`, and only on the two rows that bind V91.

It has already earned its place: it caught that `managedVersions.available[].installState.reason` reaches the same error-reason enum the row's own `managedInstallState` was just pinned away from — a second path to one declaration that a name-based reading would have missed entirely.

The 58 accepted entries are listed with the reason they are accepted, so the next person inherits a measurement instead of a search.

The second commit is that guard being fixed against itself. As first written it recorded one representative path per enum, which would have stayed green when a *new* field was wired to an enum already on the list — the exact regression it exists to catch. It also deduplicated the walk by object identity, which hid any schema reached twice. Recording every path and replacing dedup with a depth ceiling took the set 29 → 54 → 58 — the two shortcuts were hiding **half** the answer — and surfaced `auth.status` on the row itself plus the `native` union's **fourth arm**, which carries a second catch-less `denySources`. The "one no-catch leaf" claimed in the first commit is **seven**.

## Cost

The walk is not free, and the first revision claimed it was. On a payload both the head and the 7.0 row accept (8 rows with full `nativeCapabilities`, plus a native MCP result of 6 servers × 8 tools) it measures **~2.5× the frozen parse it precedes**, ~3× for the combined call — 0.037 ms → 0.12 ms. Every object and record node allocates a spread on the way down, and the union case runs `safeParse` per arm. Bounded work on a response already parsed twice over, but a multiple, not "free".

## Known gap, named not fixed

`projectResponseWithinMajor` (host `handler.ts`) has **no `providers.list` entry**, so a caller on the canonical major but an older **minor** — 9.0 against a 9.1 canonical — bypasses the projection entirely; `downgradeCanonicalToCaller` returns the canonical value untouched for a matching major. Wiring it up means a minor-keyed case at that seam, whose own comment records how a projection written for one minor silently captured the next (`epic.getTaskContexts`, where `1.1 < 1.2` made a v1.1 caller take v1.0's projection). Deliberately out of this change rather than overlooked, and now recorded in the module docblock.

## Scope

Protocol-only; no host-side change. `@9.2` (head) is untouched and still binds live schemas **by design** — that is how the next growth attempt reddens the catalog snapshot first.

## Verification

- `protocol` host suites: **46 files, 1118 tests, all passing**
- `frozen-catalog-lines`: 38/38, **unchanged bytes** on every row
- ablation control: snapshot blind, new guard red on the exact path
- depth ceiling: deepest real path measured at **13** (`native|0|0.servers[].transport|0.env[].name`), identical on all four released rows and the head
- pre-commit: build · compile · lint · format · DCO all passed


hdkshingala:
@codex review

hdkshingala:
Fixed in f3ad8ae27 — both references now read `__tests__/frozen-lines-reach-no-unreviewed-live-enum.test.ts`. The name was written into those docblocks before it settled and never swept, so two of the five references pointed at nothing. Good catch.

One note on the red `test (desktop darwin + packaging)`: it is `vite-renderer-aliases.test.ts` timing out at 5000ms, not an assertion. That test reads only gui-app's own `src` tree and `vite.renderer.config.ts`, and this PR touches neither — it changes `protocol/src/host/` only, and its one new import is relative (`./frozen-line-projection`), so it introduces no workspace-prefixed specifier for that test to check. It passes locally on this branch in 641ms against the 5000ms budget. Attributing it to runner slowness; this push re-runs it.

@codex review

hdkshingala:
@coderabbitai re the outside-diff finding — **"Project the row before the live pre-pass"** (`provider-schemas.ts:4402-4425`): I checked it and it's a false positive. Pinned as a test rather than just answered here, because the reading error is an easy one to make.

The premise is that the pre-pass parses `sharedWithProviders` through `z.array(providerIdSchema).catch([])` and so hands the projection an already-emptied array. But that pre-pass parses with the **live** schema, and *"newer than v7.0"* is not *"unknown to live"*. `antigravity` is a member of the live enum, so the live `.catch([])` never fires — only the **frozen** v7.0 enum refuses it, and by then the projection is the thing doing the refusing.

It's inert by construction, not by luck: `parseProviderStateWithEnabledProfiles` shares its schema with the head parse in `handler.ts` that already gated the value. Anything reaching the bridge has, by definition, satisfied the live schema.

Measured:

```
HEAD  ["claude-code","antigravity"]   both accepted
V70   ["claude-code"]                 pre-pass + projection
V80   ["claude-code"]                 no pre-pass — the control
```

v8.0 is the control precisely because it has no pre-pass. If the pre-pass were the hazard the two lines would disagree; they don't.

`0214580c4` asserts all three, plus the precondition the argument rests on (that live accepts both ids), so a future edit narrowing the live enum fails it loudly instead of letting it pass for a new reason.

Worth noting the existing test `drops only the newer id for a v7.0 peer` already covered this path end-to-end through `downgradeResponseAcrossMajors` — it just didn't *name* the pre-pass, which is what left room for the reading.

---

Separately: your last review was on `f3ad8ae27`, and the CodeRabbit check has been reporting **"Review rate limited"** since. Two commits have landed since then — `a689218df` (two behaviour fixes to the projection: union arms now score the *parsed* result rather than the projected input, and the walk never empties a non-empty array) and `0214580c4`. A fresh review of `0214580c4` would be useful before this merges.

hdkshingala:
@coderabbitai approve

PR #2033, opened 2026-09-20
feat(protocol,clients): user record major 2 with APPLE, negotiated by the clients

hdkshingala:
## Summary

Groundwork for Sign in with Apple. The auth `providerType` enum is closed, so a new `APPLE` value would fail `safeParse` in every released client and read as a credential rejection. This adds the value behind a record major and teaches the clients to negotiate it.

**Protocol (`protocol/src/auth`)**
- Major 2 of `user`, `authenticated-user-response`, `provider-login-response`, `exchange-token-response` and `list-all-mcp-servers-response` accepts `APPLE`. `organization` and `legacy-authenticated-user-response` stay on major 1.
- One real downgrade bridge, owned by `user`: `APPLE` maps to `EMAIL`. The envelope delegates to it. Upgrades are identity.
- A freeze test pins the fingerprint of every retained major-1 contract, so major 1 cannot drift or alias the major-2 enum by accident. A diff in that file is a major-1 breach, never a snapshot refresh.
- `ProviderType` is now derived from the latest `User`.

**Clients (`clients/shared`, `clients/gui-app`)**
- Identity validation calls `GET /api/v3/user/negotiated`, advertising `x-traycer-user-record-majors: 1,2`, and parses at the major named by `x-traycer-user-record-version`.
- Recovery onto the frozen `GET /api/v3/user` is private to endpoint selection: only on a 404 or a 2xx whose version label is missing, malformed or unsupported. Never on 401/403. Both requests share one abort and deadline, and the pair is one unit under `withAuthNetworkRetry`. The frozen route keeps its own classification, so a genuine 404 still ends as `rejected`.
- A missing version header can no longer reach the schema-failure branch, which reads as a credential rejection and can spend a refresh token.
- The provisional session snapshot stays on major 1 (the released reader rejects any other stamp before parsing, and the key is shared across desktop versions): downgraded before persisting, upgraded on read.
- Account deletion form copy mentions Apple.
- Existing fetch fakes keyed on `/api/v3/user` were repaired for the new first request.

No caller of this validation runs in a browser context (electron-main, Node CLI, Capacitor native HTTP), so the custom response header needs no CORS exposure.

## Related issue

None.

## Checklist

- [x] Pre-commit static checks pass (hook ran on every commit; `bun run compile` clean with the Nx cache skipped)


hdkshingala:
@codex review

hdkshingala:
@codex review

